PR #879, opened 2026-01-09
fix: key storage locally

yu2C:
Avoid storing keys in the database to prevent exposure and keep them in local storage instead.

### DIFF
--- a/internal/keystore/local/local.go
+++ b/internal/keystore/local/local.go
@@ -143,6 +143,10 @@ func (l *LocalKeyStore) Import(t keystore.KeyType, seed []byte) (keystore.KeyPai
 	}
 
 	pubHex := hex.EncodeToString(kp.PublicKey())
+	if _, exists := keys[pubHex]; exists {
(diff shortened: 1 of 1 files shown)


internal/keystore/local/local.go
```diff
@@ -246,3 +257,25 @@ func (l *LocalKeyStore) Delete(t keystore.KeyType, pubKey []byte) error {
 	}
 	return nil
 }
+
+// ImportValidatorKeysFromSeed imports both Ed25519 and Bandersnatch keys
```
YCC3741: In `ImportValidatorKeysFromSeed()`, if the second import fails (line 276), the first key (line 271) is already saved. This means we could end up with only one key stored instead of both.

Is this okay for your use case? 

If not, we could add simple rollback:
```go
if _, err := l.Import(keystore.KeyTypeBandersnatch, bandersnatchSecretSeed); err != nil {   
    _ = l.Delete(keystore.KeyTypeEd25519, edKP.PublicKey())    
    return fmt.Errorf("failed to import Bandersnatch key: %w", err)
}
```

PR #885, opened 2026-01-16
Fix/0.7.2 pvm bugs

TwEricShen:
Since folderwise is removed, fuzzer cmd:
`go run ./cmd/fuzz test_folder /tmp/jam_target.sock pkg/test_data/jam-conformance/fuzz-reports/0.7.2/traces/ `


yu2C (approved):
LGTM. it passed `conformance test` now

PR #911, opened 2026-03-01
feat: add node rpc & subscription feature

weigen393:


internal/eventbus/types.go
```diff
@@ -0,0 +1,50 @@
...
+}
+
+type ChainSubscriptionUpdate struct {
+	HeaderHash string      `json:"header_hash"`
+	Slot       uint64      `json:"slot"`
+	Value      interface{} `json:"value"`
+}
+
+type ServiceEvent struct {
+	ServiceID  string      `json:"service_id"`
```
yu2C, line 35: Use number? u32

internal/rpc/subscription_test.go
```diff
@@ -0,0 +1,307 @@
...
+	}
+
+	var subID1 uint64
+	select {
+	case resp1 := <-responseCh:
+		if resp1.Error != nil {
+			t.Fatalf("First subscribe request returned error: %v", resp1.Error)
+		}
+		subID1 = uint64(resp1.Result.(float64))
+		t.Logf("First subscription ID: %s", subID1)
```
yu2C, line 250: %d

internal/rpc/subscription_test.go
```diff
@@ -0,0 +1,307 @@
...
+	}
+
+	var subID2 uint64
+	select {
+	case resp2 := <-responseCh:
+		if resp2.Error != nil {
+			t.Fatalf("Second subscribe request returned error: %v", resp2.Error)
+		}
+		subID2 = uint64(resp2.Result.(float64))
+		t.Logf("Second subscription ID: %s", subID2)
```
yu2C, line 268: %d

internal/rpc/subscription.go
```diff
@@ -0,0 +1,184 @@
...
+	if sub.Transform != nil {
+		result = sub.Transform(event)
+	}
+
+	return &JSONRPCNotification{
+		JSONRPC: "2.0",
+		Method:  sub.MethodName,
+		Params: NotificationParams{
+			Subscription: sub.ID,
+			Result:       result,
```
yu2C, line 158: Do we need error handling?

cmd/node/main.go
```diff
@@ -72,51 +77,66 @@ func init() {
 func node(ctx context.Context, cmd *cli.Command) error {
 	config.InitConfig(configPath, mode)
 	SetupJAMProtocol(chainPath)
+
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
```
yu2C, line 82: **Blocker — RPC server lifecycle vs process exit**

The RPC server is started in a goroutine with `go func() { ... rpcServer.Start(ctx) }`, while the main goroutine returns immediately after `cancel()` (triggered by signal) without waiting for the server goroutine to finish shutting down.

`context.Cancel` + `http.Server.Shutdown` (inside `Start`) is asynchronous with respect to the `node()` return. The process can exit while connections are still open or `ListenAndServe` / `Shutdown` is still in progress, leading to unclean termination and lost in-flight work.

**Suggestion:** use a `sync.WaitGroup` (or a dedicated `done` channel) so that after `cancel()` the main path waits until `Start` has returned (or until a bounded timeout), matching the 5s shutdown in `Start`.

cmd/node/main.go
```diff
@@ -72,51 +77,66 @@ func init() {
...
+
+	go func() {
+		chainState := blockchain.GetInstance()
+		eventBus := eventbus.NewEventBus()
```
yu2C, line 86: **Architecture — EventBus must be a shared process dependency**

`eventbus.NewEventBus()` is created inside the RPC goroutine and only passed to `NewRPCServer`. The rest of the node (e.g. blockchain, sync, block import) has no way to `Publish` on the same bus, so real chain events cannot drive RPC subscribers without duplicating the bus or resorting to polling (e.g. `ChainWatcher` only).

**Suggestion:** construct a single `EventBus` (or `EventPublisher` / `EventSubscriber` interfaces) at the node composition root (`node()` or immediately after `SetupJAMProtocol`), and inject it into both the RPC server and any component that should emit JIP-2–relevant events, so `Publish` and `Subscribe` use the same instance.

PR #923, opened 2026-03-21
refactor: update TicketAttempt codec methods to use compact encoding

yu2C:


YCC3741:
**TicketAttempt decode after switching to compact**

`DecodeLength()` returns a `uint64`, and `*t = TicketAttempt(val)` narrows to the underlying `U8`. For `val > 255`, Go truncates silently (e.g. `256 → 0`), which could not happen with the previous fixed single-byte read.

Suggest adding an explicit guard before the assignment, e.g. return a decode error if `val > math.MaxUint8`, so we keep GP-style compact on the wire but reject values that do not fit the `U8` type. Stricter rules (e.g. vs `TicketsPerValidator`) can stay in safrole / validation layers.

YCC3741 (changes requested):
As above

YCC3741 (approved):

PR #925, opened 2026-03-27
feat: Validator grid and UP/CE networking infrastructure

yu2C:
## Summary

This PR adds **validator grid topology and neighbor computation** aligned with JAMNP, and strengthens **QUIC connection handling / stream dispatch** and **CE handler wire consistency**. It also includes small **API, naming, and index-safety** cleanups so follow-up work (UP 0, automatic dialing, epoch transitions) can land cleanly.

## Motivation

- JAMNP ties connectivity and UP 0 peers to **grid neighbors** and **previous / current / next** epoch validator sets; the codebase lacked a reusable grid / manager layer.
- **EventBus** used inconsistent keys for `Subscribe` vs `Publish`, so handlers never ran.
- **ConnectionManager** was not concurrency-safe (r/w on a map without a mutex).
- **CE** handling mixed **1-byte stream kind** with an extra **4-byte length** wrapper in one path, which could desynchronize reads inside handlers.
- **Peer** needed an accept loop, TLS-backed Ed25519 peer identity, **1-byte stream kind** dispatch, and **Broadcast** that writes the kind byte before the payload.
- **PeerSet** needed multi-index lookups (key / address / validator index) and must reject malformed Ed25519 key lengths to avoid corrupt indices.
- **GridMapper** field names were updated to **`Previous` / `Current` / `Next`** for clearer alignment with the spec.

### P0 / foundational fixes

| File | Change |
|------|--------|
| `internal/networking/quic/event_bus.go` | `handlers` as `map[EventType][]Handler`; `Publish(ctx, eventType, event)` |
| `internal/networking/quic/connection.go` | Mutex on `ConnectionManager`; `Add` / `Remove` / `All` |

### Validator grid and dial addresses

| File | Change |
|------|--------|
| `internal/networking/validator/grid.go` | `GridMapper` (`Previous` / `Current` / `Next`), `ComputeWidth`, same-epoch neighbors, same-index across epochs, `FindIndex` |
| `internal/networking/validator/manager.go` | `ValidatorManager`, `PreferredInitiator`, metadata → IPv6 + little-endian port |

### Peer / stream dispatch

| File | Change |
|------|--------|
| `internal/networking/quic/peer_set.go` | Multi-index map; `Add` returns `error`; Ed25519 key length must be 32 bytes when a key is present |
| `internal/networking/quic/peer.go` | `Start` / accept loop; validate TLS cert and extract peer key; `RegisterHandler`; `Broadcast` writes stream kind; integrates `PeerSet` |
| `internal/networking/quic/handler.go` | `HandleStreamByKind`; `HandleStream` keeps backward-compatible 1-byte read then dispatch |

### CE dispatch

| File | Change |
|------|--------|
| `internal/networking/handler/ce/ce_handler.go` | Removed incorrect outer 4-byte framing; `HandleStream(blockchain, protoID, stream)` |

### Tests

| File | Change |
|------|--------|
| `internal/networking/quic/peer_test.go` | `Ed25519Key`, `Connect` signature, broadcast wire includes kind byte |
| `internal/networking/handler/ce/ce128_test.go`, `ce129_test.go` | Updated `Connect` calls |

## Validation

```bash
go test ./internal/networking/quic ./internal/networking/validator ./internal/networking/handler/ce
```

Expected: `quic` and `handler/ce` pass; `validator` currently has no test files.

## Out of scope (follow-ups)

- Full UP 0 block announcement state machine (handshake loop, announcement loop, duplicate UP stream resolution).
- Epoch-transition connectivity delay rules from JAMNP.
- Preferred-initiator 5s fallback and dialing **all** prev/current/next validators automatically.
- Restoring explicit **`PeerRole` / builder ALPN** on `Connect` if product wiring needs it (small follow-up PR).
- Replacing `log` in networking with the shared `logger` package for consistent levels and configuration.

### DIFF
--- a/internal/networking/quic/peer.go
+++ b/internal/networking/quic/peer.go
@@ -4,9 +4,12 @@ import (
 	"context"
 	"crypto/ed25519"
 	"crypto/tls"
+	"fmt"
 	"log"
 	"net"
+	"sync"
 
+	networkcert "github.com/New-JAMneration/JAM-Protocol/internal/networking/cert"
 	"github.com/New-JAMneration/JAM-Protocol/internal/types"
 	"github.com/quic-go/quic-go"
 )
@@ -33,14 +36,22 @@ type PeerConfig struct {
 	CEHandler     CEHandler
 }
 
+type StreamHandlerFunc func(ctx context.Context, stream *Stream, peerKey ed25519.PublicKey) error
+
 type Peer struct {
-	publicKey   ed25519.PublicKey
-	Listener    *Listener
-	tlsConfig   *tls.Config
-	quicConfig  *quic.Config
-	connManager *ConnectionManager
-	CEHandler   CEHandler
-	UPHandler   UPHandler
+	Ed25519Key     ed25519.PublicKey
+	ValidatorIndex *uint16
+	Listener       *Listener
+	tlsConfig      *tls.Config
+	quicConfig     *quic.Config
+	connManager    *ConnectionManager
+	peerSet        *PeerSet
+	ctx            context.Context
+	cancel         context.CancelFunc
+	handlerMu      sync.RWMutex
+	handlers       map[byte]StreamHandlerFunc
+	CEHandler      CEHandler
+	UPHandler      UPHandler
 	// Sync-related fields
 	Best      *HeadInfo `json:"best"`      // Optional best block header
 	Finalized HeadInfo  `json:"finalized"` // Finalized block header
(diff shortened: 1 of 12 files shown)


internal/networking/validator/grid.go
```diff
@@ -0,0 +1,86 @@
...
+		result = append(result, g.Previous[index])
+	}
+	if index < len(g.Next) {
+		result = append(result, g.Next[index])
+	}
+
+	return result
+}
+
+func (g *GridMapper) IsNeighborInEpoch(a, b int) bool {
```
Terryhung: Should `IsNeighbor(key)` consider peers whose key exists only in Previous/Next (same validator index), since JAMNP-S counts same-index-across-epochs as grid neighbours, or is Current-only lookup?

internal/networking/quic/peer.go
```diff
@@ -79,27 +92,108 @@ func NewPeer(config PeerConfig) (*Peer, error) {
...
 	}
 
-	return conn.(*Connection), nil
+	if err := handler(p.ctx, stream, peerKey); err != nil {
+		log.Println("stream handler error:", err)
+		_ = stream.Close()
+	}
 }
 
 func (p *Peer) Broadcast(kind string, message interface{}) {
```
Terryhung: Should Broadcast follow JAMNP-S message framing (4-byte LE length after the stream kind byte)?

Issue #824, opened 2025-12-06
refactor: replace repeated type hashing procedures to `Hash()` method


shunsukew:
Currently, hashing types (for e.g. `Header`) doesn't have easy to use interface for hashing, all developers need to follow same procedures many times across codebase. This approach has 2 possible downsides.

1. Developers always need to know the specification of type hashing, which disable them to focus on bussiness logic, and more error prone.
2. Everytime, creating new encoder with new allocated buffer, which is memory inefficient.

```
encoder := types.NewEncoder()
encodedHeader, err := encoder.Encode(&header)
if err != nil {
	return types.HeaderHash{}, err
}
headerHash := types.HeaderHash(Blake2bHash(encodedHeader)), nil
```

All types that need to be hashed can implement `Hash()` method, so that developers can easily call it without caring of hashing specification.

Proposed approach is using `sync.Pool` https://pkg.go.dev/sync#Pool.

```
var encoderPool = sync.Pool{
    New: func() any {
        return &Encoder{
            buf:            new(bytes.Buffer),
            HashSegmentMap: nil,
        }
    },
}

func (h *Header) Hash() (HeaderHash, error) {
	encoder := encoderPool.Get().(*Encoder)
	defer encoderPool.Put(encoder)

	encoded, err := encoder.Encode(h)
	if err != nil {
		return HeaderHash{}, err
	}

	return HeaderHash(blake2b.Sum256(encoded)), nil
}
```

PR #926, opened 2026-03-29
refactor: Add Encoder Pool by sync.Pool to Reduce Repeated Allocation GC Cost

HanaYukii:
## Summary

Issue: #824.

The codebase has 128 occurrences of the same `NewEncoder → Encode → Blake2bHash` pattern across 44 files. Each call allocates a new `Encoder` and buffer, then discards them for GC. Under high-frequency calls this creates unnecessary memory pressure.

### Changes

- Add `sync.Pool` for `Encoder` instances (`GetEncoder`/`PutEncoder`) to reduce GC pressure
- Add `hash.HashEncode(v Encodable)` — single-call replacement for the repeated `NewEncoder → Encode → Blake2bHash` pattern
- Refactor `ComputeBlockHeaderHash` to use `HashEncode` internally (no signature change)
- Keep `NewEncoder()` for backward compatibility, marked as deprecated

### Before / After

```go
// Before
encoder := types.NewEncoder()
encoded, err := encoder.Encode(&header)
h := hash.Blake2bHash(encoded)

// After
h, err := hash.HashEncode(&header)
```
### Benchmark

Measured via `go test -bench=BenchmarkHashEncode -benchmem -count=3` on `Header{}`:

BenchmarkHashEncode_Pooled-24 ~5.6k ns/op 1272 B/op 20 allocs/op
BenchmarkHashEncode_NoPool-24 ~8.3k ns/op 2296 B/op 26 allocs/op

~35% faster, 45% less memory, 23% fewer allocations.

Pooled:    ~5.6k ns/op   1272 B/op   20 allocs/op
No pool:   ~8.3k ns/op   2296 B/op   26 allocs/op

~35% faster, 45% less memory, 23% fewer allocations.

### DIFF
--- a/internal/types/encoder.go
+++ b/internal/types/encoder.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"math/bits"
+	"sync"
 )
 
 type Encoder struct {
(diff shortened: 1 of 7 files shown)


YCC3741 (changes requested):
### 1. Run full test suite with FluffyLabs JAM Testing Docker

Before merging, please run the complete test suite using the [FluffyLabs JAM Testing](https://github.com/FluffyLabs/jam-testing) Docker setup to verify that the encoder pool refactoring does not introduce any regressions.

### 2. EncodeMany pool safety - missing bytes.Clone

`Encode` returns `bytes.Clone(e.buf.Bytes())`, which is safe for pooled reuse. However, `EncodeMany` returns `e.buf.Bytes()` directly **without cloning**.

With the introduction of `sync.Pool` via `GetEncoder`/`PutEncoder`, this becomes a latent bug: if a caller uses `GetEncoder` -> `EncodeMany` -> `PutEncoder`, the returned slice aliases the pooled buffer. Once another goroutine retrieves the same encoder from the pool, the buffer will be overwritten, causing silent data corruption.

**Current code (internal/types/encoder.go):**

```go
func (e *Encoder) EncodeMany(vs ...any) ([]byte, error) {
	cLog(Cyan, "Encoding")
	e.buf.Reset()

	for _, v := range vs {
		if err := e.encodeStruct(v); err != nil {
			return nil, err
		}
	}

	return e.buf.Bytes(), nil  // not safe with pooled encoders
}
```

**Suggested fix - either:**

**(a) Change EncodeMany to clone, consistent with Encode (strongly preferred):**

```go
return bytes.Clone(e.buf.Bytes()), nil
```

**(b) Or at minimum, add a doc comment warning callers:**

```go
// EncodeMany encodes multiple values into a single byte slice.
// WARNING: The returned slice references the encoder's internal buffer.
// Do NOT use with GetEncoder/PutEncoder - use Encode instead, which
// returns a cloned copy.
```

Option (a) is strongly preferred for consistency and safety.

YCC3741 (approved):
LGTM

PR #932, opened 2026-04-07
feat: update TLS configuration handling to use pre-generated certificates

yu2C:
Waiting #925 , and should be merged after that

### Summary

- Refactor TLS setup in `internal/networking/cert` to work cleanly with pre-generated certificates and shared helpers (e.g. building server/client `tls.Config` from existing key material).
- Add Ed25519 / TLS test coverage (generator_test).
- Wire QUIC Peer (and config) to use the updated certificate / key path so connections use the same TLS story.
- Touch CE ce128 / ce129 tests only as needed for the new helpers or types (no protocol behavior change intended there).

### DIFF
--- a/internal/networking/handler/ce/ce128_test.go
+++ b/internal/networking/handler/ce/ce128_test.go
@@ -402,16 +402,16 @@ func TestCE128RequestWithPeer(t *testing.T) {
 
 	ceHandler.RegisterCEHandler(128, HandleBlockRequestStream)
 
-	publicKey, _, err := ed25519.GenerateKey(rand.Reader)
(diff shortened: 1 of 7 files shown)


internal/networking/handler/ce/ce128_test.go
Terryhung: (nit) non-blocking: maybe check the err is not nil first.

Terryhung (approved):

internal/networking/handler/ce/ce128_test.go
yu2C: updated

YCC3741 (approved):

PR #933, opened 2026-04-12
fix: set ValidatorID in ComputeInitialAuditAssignment + Stage 1 tests

HanaYukii:
## Bug Fix (#935)

`ComputeInitialAuditAssignment()` did not set `ValidatorID` when building `AuditReport` (defaulted to 0).

This field is read by `UpdatePositiveJudgersFromAudit()` and `IsWorkReportAudited()` to match who was assigned vs who judged. Wrong ID causes Rule 1 to always fail, forcing fallback to supermajority (Rule 2).

Fix: add `ValidatorID: validatorIndex`. Extract `buildInitialAuditAssignmentFromCoreOrder()` for testability.

## Tests (4 new)

| Code does | Test checks | GP |
|-----------|------------|:--:|
| Collect candidates from ρ (assignments) and W (available reports) | Reports in ρ but not in W get filtered out | 17.2 |
| Build initial assignment from shuffled core order | ValidatorID is set on each entry (regression) | 17.5 |
| Track which validators are assigned to each report (assignmentMap) | ValidatorID propagates from computed assignments | 17.13 |
| Sign audit announcement with Ed25519 (BuildAnnouncement) | Reconstructed signing context matches signature | 17.9 |

Also fix pre-existing `ServiceId` → `ServiceID` typo so the package compiles.

See #931 for full test plan (Stage 2-4 in separate branches).

## Tested

```bash
go test -mod=mod -v -run "TestCollectCandidates_|TestAssignmentMap_|TestAnnouncement_|TestInitialAssignment_" ./internal/auditing/  # all 4 pass
go test -mod=mod ./internal/auditing/  # full package pass
```

Tested in WSL. Windows has CGO linker issue with Rust FFI — use WSL or CI.

### DIFF
--- a/internal/auditing/auditing.go
+++ b/internal/auditing/auditing.go
@@ -124,7 +124,14 @@ func ComputeInitialAuditAssignment(Q []*types.WorkReport, validatorIndex types.V
 	}
 	shuffled := shuffle.Shuffle(coreIndices, types.OpaqueHash(vrfOutput))
 
-	// Step 4: Select top 10 assigned reports (17.5)
+	return buildInitialAuditAssignmentFromCoreOrder(Q, validatorIndex, shuffled), nil
+}
+
+func buildInitialAuditAssignmentFromCoreOrder(
+	Q []*types.WorkReport,
+	validatorIndex types.ValidatorIndex,
+	shuffled []types.U32,
+) []types.AuditReport {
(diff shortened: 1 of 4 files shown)


HanaYukii:
Reopening as clean PR.

PR #934, opened 2026-04-12
fix: set ValidatorID in ComputeInitialAuditAssignment

HanaYukii:
## What

`ComputeInitialAuditAssignment()` builds `AuditReport` structs but did not set `ValidatorID` (defaults to 0).

This field is read later by `UpdatePositiveJudgersFromAudit()` and `IsWorkReportAudited()` to match who was assigned vs who judged. With the wrong ID, Rule 1 lookup always misses, so audit can only pass through supermajority (Rule 2).

## Fix

- Add `ValidatorID: validatorIndex` when building each `AuditReport`
- Extract `buildInitialAuditAssignmentFromCoreOrder()` helper for testability
- Include regression test (`validatorid_regression_test.go`)

Tracked by #935.


HanaYukii:
Merged into #933 as a single PR.

Issue #935, opened 2026-04-12
fix: ComputeInitialAuditAssignment missing ValidatorID


HanaYukii:
`ComputeInitialAuditAssignment()` builds `AuditReport` structs but does not set `ValidatorID` (defaults to 0).

This field is read later by `UpdatePositiveJudgersFromAudit()` and `IsWorkReportAudited()` to match who was assigned vs who judged. With the wrong ID, Rule 1 lookup always misses, so audit can only pass through supermajority (Rule 2).

Fixed in #934.

HanaYukii:
Fixed by #936 (merged 2026-04-26). Note body references #934, but the actual fix landed in #936.

PR #936, opened 2026-04-12
fix/test: set ValidatorID bug + Stage 1 tests

HanaYukii:
## Bug Fix (#935)

`ComputeInitialAuditAssignment()` did not set `ValidatorID` when building `AuditReport` (defaulted to 0).

This field is read by `UpdatePositiveJudgersFromAudit()` and `IsWorkReportAudited()` to match who was assigned vs who judged. Wrong ID causes Rule 1 to always fail, forcing fallback to supermajority (Rule 2).

Fix: add `ValidatorID: validatorIndex`. Extract `buildInitialAuditAssignmentFromCoreOrder()` for testability.

## Tests (3 new)

| Code does | Test checks | GP |
|-----------|------------|:--:|
| Pick which reports to audit — take each core's assigned report (ρ), keep only those available locally (W) | 5 cores: 2 available kept, 2 unavailable filtered, 1 unassigned nil | 17.2 |
| Walk cores in VRF-shuffled order, skip nil, pick non-nil reports as audit assignments | ValidatorID set on each entry (regression #935), nil cores skipped, result order follows shuffle | 17.5 |
| Sign audit announcement with Ed25519 (BuildAnnouncement) | Reconstructed signing context matches signature | 17.9 |

Also fix pre-existing `ServiceId` → `ServiceID` typo so the package compiles.

See #931 for full test plan (Stage 2-4 in separate branches).

## Tested

\`\`\`bash
go test -mod=mod -v -run "TestCollectCandidates|TestAnnouncement_|TestInitialAssignment_" ./internal/auditing/  # all 3 pass
go test -mod=mod ./internal/auditing/  # full package pass
\`\`\`

Tested in WSL. Windows has CGO linker issue with Rust FFI — use WSL or CI.

### DIFF
--- a/internal/auditing/stage1_test.go
+++ b/internal/auditing/stage1_test.go
@@ -1,0 +1,217 @@
+// Stage 1 tests — pure audit judgment logic that does not depend on
+// networking, VRF, or PVM execution. See docs/ce-audit-test-plan.md
+// and issue #931 for the full staged test plan.
+package auditing
+
+import (
+	"crypto/ed25519"
+	"testing"
+
+	"github.com/New-JAMneration/JAM-Protocol/internal/blockchain"
+	"github.com/New-JAMneration/JAM-Protocol/internal/types"
+	"github.com/New-JAMneration/JAM-Protocol/internal/utilities"
+	"github.com/New-JAMneration/JAM-Protocol/internal/utilities/hash"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// ---------------------------------------------------------------------------
+// Stage 1 helpers
+// ---------------------------------------------------------------------------
+
+// setupDeterministicAuditChain sets tiny mode, resets blockchain singleton,
+// and returns a fixed Ed25519 private key for signing tests.
+func setupDeterministicAuditChain(t *testing.T) ed25519.PrivateKey {
+	t.Helper()
+
+	types.SetTinyMode()
+	blockchain.ResetInstance()
+	t.Cleanup(blockchain.ResetInstance)
+
+	cs := blockchain.GetInstance()
+	var entropy types.BandersnatchVrfSignature
+	entropy[0] = 1
+
+	cs.GetProcessingBlockPointer().SetHeader(types.Header{
+		Slot:          100,
+		AuthorIndex:   0,
+		EntropySource: entropy,
+	})
+
+	seed := make([]byte, ed25519.SeedSize)
+	seed[0] = 42
+	return ed25519.NewKeyFromSeed(seed)
+}
+
+// makeDetailedWorkReport builds a WorkReport with all fields populated using
+// unique values derived from id, so map lookups stay easy to trace in tests.
+func makeDetailedWorkReport(id byte, coreID types.CoreIndex) types.WorkReport {
+	var hashValue types.WorkPackageHash
+	hashValue[0] = id
+
+	var erasureRoot types.ErasureRoot
+	erasureRoot[0] = id + 10
+
+	var exportsRoot types.ExportsRoot
+	exportsRoot[0] = id + 20
+
+	var codeHash types.OpaqueHash
+	codeHash[0] = id + 30
+
+	var payloadHash types.OpaqueHash
+	payloadHash[0] = id + 40
+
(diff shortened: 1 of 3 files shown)


internal/auditing/stage1_test.go
```diff
@@ -0,0 +1,217 @@
...
+	cs.GetProcessingBlockPointer().SetHeader(types.Header{
+		Slot:          100,
+		AuthorIndex:   0,
+		EntropySource: entropy,
+	})
+
+	seed := make([]byte, ed25519.SeedSize)
+	seed[0] = 42
+	return ed25519.NewKeyFromSeed(seed)
+}
```
YCC3741, line 70: **SUGGESTION — `setupDeterministicAuditChain` violates SRP**

This helper does four unrelated things:
1. Switches to tiny mode
2. Resets the blockchain singleton
3. Sets a header (entropy/slot/authorIndex)
4. Generates and returns an Ed25519 private key

However:
- `TestCollectCandidates` and `TestInitialAssignment_SetsValidatorID` do **not** need the Ed25519 key (they don't go through `BuildAnnouncement` / VRF).
- They also don't need the header — only `TestAnnouncement_SignContext` actually reads from `blockchain.GetInstance().GetProcessingBlockPointer().GetHeader()`.

Suggest splitting into:
- `setupChainState(t)` — common: tiny mode + `ResetInstance` (used by all three tests)
- `setupForAnnouncement(t)` — adds header setup and returns the Ed25519 key (used only by `TestAnnouncement_SignContext`)

This makes each test's preconditions explicit and avoids the misleading 'every test has a header and key' implication.

internal/auditing/stage1_test.go
```diff
@@ -0,0 +1,217 @@
...
+// CollectAuditReportCandidates (GP 17.1-17.2)
+// ---------------------------------------------------------------------------
+
+// GP 17.2: Q keeps reports that are both assigned (ρ) and available (W), filters out the rest.
+func TestCollectCandidates(t *testing.T) {
+	setupDeterministicAuditChain(t)
+
+	// Use 5 cores so we can test a mix of assigned/available/filtered.
+	types.CoresCount = 5
+	t.Cleanup(func() { types.SetTinyMode() })
```
YCC3741: **SUGGESTION — Direct mutation of global `types.CoresCount` is fragile**

```go
types.CoresCount = 5
t.Cleanup(func() { types.SetTinyMode() })
```

This pattern is repeated in both `TestCollectCandidates` (L115-116) and `TestInitialAssignment_SetsValidatorID` (L150-151).

Two concerns:
1. **Parallelism hazard**: if anyone in this package adds `t.Parallel()` later, these tests will race against any other test that reads/writes `types.CoresCount`.
2. **Cleanup ordering is non-obvious**: `setupDeterministicAuditChain` already registers `t.Cleanup(blockchain.ResetInstance)`, and the test adds its own `t.Cleanup(SetTinyMode)`. LIFO order happens to work here, but a future reader has to reason about it.

Two options (either is fine):
- **Minimum**: add a top-of-file comment stating these tests must not run in parallel with each other or other tests touching global state.
- **Better**: extend the helper with an optional cores parameter (e.g. `setupChainState(t, withCores(5))`) and let the helper own both the mutation and the cleanup, so each test stops repeating the same two lines.

YCC3741 (changes requested):
Two test-structure issues that should be addressed before merge. The core bug fix in `auditing.go` looks good — the missing `ValidatorID` is a real regression and the change is minimal and correct. Comments below are about `stage1_test.go` only.

internal/auditing/stage1_test.go
HanaYukii: Addressed. I split the old helper into setupChainState and setupForAnnouncement, so only the signing test sets up header + Ed25519 key and the other tests keep the minimal shared state setup.

internal/auditing/stage1_test.go
HanaYukii: Addressed. I moved the CoresCount override and cleanup into setupChainState(withCores(...)), and added a file-level note that these tests must not call t.Parallel() because they mutate package-global state. Re-ran go test -mod=mod ./internal/auditing in WSL after the change.

yu2C (approved):

PR #939, opened 2026-04-26
fix/test: Stage 2 BuildJudgements key bug + tranche tests

HanaYukii:
## Fixes

- `BuildJudgements()` was passing `Kappa[v].Ed25519` (32B public key) to `ed25519.Sign()`, which needs a 64B private key. Calls panic with `bad private key length: 32`. Take `validatorPrivKey ed25519.PrivateKey` from the caller, matching `BuildAnnouncement`. (GP §17.17 `S_κ[v]ᵉ` means sign with the validator's signing key, not look up the public key entry in chain state.)
- `GetTranchIndex()` clamps to 0 when `T < P·Ht` would underflow u64. GP §17.7 implies this never happens but the guard is cheap.

## Tests (4 in fresh `internal/auditing/stage2_test.go`)

| Function | Cases | GP |
|---|---|:-:|
| `BuildJudgements` | rebuild ⟨Xe ⌢ H(w)⟩, verify against derived pubkey (regression for fix above); empty input does not panic | 17.17 |
| `GetTranchIndex` | n = (T − P·Ht) / A, sandwich tBefore/tAfter for boundary crossings; future slot clamps to 0 instead of u64 wrap | 17.8 |

`WaitNextTranche` is not retested here — main's `auditing_test.go` already has `TestWaitNextTranche_{NormalExpiry, ContextCancel, AlreadyPastDeadline, Tranche2, ContextDeadlineExceeded}` from #910, which covers the same surface.

`GetJudgement` fetch / process error paths are not here either, also covered on main by #910.

See #931 for the audit test plan.

## Tested

\`\`\`bash
go test -mod=mod -v -run "TestStage2_" ./internal/auditing/  # 4 pass
go test -mod=mod ./internal/auditing/                        # full pkg pass
\`\`\`

WSL only. Windows CGO / Rust FFI breaks linking elsewhere in the repo.

### DIFF
--- a/internal/auditing/auditing.go
+++ b/internal/auditing/auditing.go
@@ -152,13 +152,22 @@ func buildInitialAuditAssignmentFromCoreOrder(
 }
 
 // (17.8) let n = (T − P ⋅ Ht) / A
-// GetTranchIndex computes tranche index from wall-clock time and block slot
+// GetTranchIndex computes tranche index from wall-clock time and block slot.
+//
+// GP §17.7 implies T ≥ P·Ht (a tranche cannot occur before its block's slot).
+// If a caller hands us a future slot anyway (clock skew, malformed header,
+// future-slot lookahead), the unguarded subtraction would underflow u64 to
+// a huge bogus tranche index. Clamp to 0 in that case so downstream
+// arithmetic stays sane.
 func GetTranchIndex() types.U64 {
 	T := types.U64(header.GetCurrentTimeInSecond())                                 // T current time (seconds)
 	Ht := types.U64(blockchain.GetInstance().GetProcessingBlockPointer().GetSlot()) // Ht slot number from block header
 	P := types.U64(types.SlotPeriod)                                                // P: seconds per slot
 	A := types.U64(types.TranchePeriod)                                             // A: seconds per tranche
-	n := (T - P*Ht) / A                                                             // n = (T - P ⋅ Ht) / A
+	if T < P*Ht {
+		return 0
+	}
+	n := (T - P*Ht) / A // n = (T - P ⋅ Ht) / A
 	return n
 }
 
@@ -376,12 +385,18 @@ func EvaluateReport(
 	return nil // ⊥ — evaluation failed
 }*/
 
-// (17.18) n = {Sκ[v]e (Xe(w) ⌢ H(w)) S (c, w) ∈ an}
+// (GP §17.17, v0.7.2; older revisions §17.18) jn = {S_κ[v]ᵉ ⟨Xe(w) ⌢ H(w)⟩ | (c, w) ∈ aₙ}
+//
+// Each judgment is signed with the validator's Ed25519 private key. The chain-state
+// entry Kappa[v].Ed25519 is the public key (Ed25519Public, 32B); passing it to
+// ed25519.Sign panics with "bad private key length: 32". The private key must
+// be supplied by the caller (same pattern as BuildAnnouncement).
 func BuildJudgements(
 	tranche types.U8,
 	auditReports []types.AuditReport, // (c, w) ∈ aₙ
 	hashFunc func(types.ByteSequence) types.OpaqueHash,
 	validator_index types.ValidatorIndex,
+	validatorPrivKey ed25519.PrivateKey, // κ[v]ᵉ: Ed25519 private key
 ) []types.AuditReport {
 	for index, audit := range auditReports {
 		report := audit.Report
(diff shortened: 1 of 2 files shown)


YCC3741 (commented):
## Summary

Solid, focused fix/test PR. The `BuildJudgements` key bug fix is correct — passing the
32-byte `Ed25519` public key to `ed25519.Sign` panics with `bad private key length: 32`,
and taking the private key from the caller (matching `BuildAnnouncement`) is the right
approach. The `GetTranchIndex` u64-underflow clamp is a cheap, sensible guard, and the new
tests are genuine regression tests. No blocking issues.

cc @yu2C — could you also take a look when you get a chance?

---

### `internal/auditing/auditing.go`

#### :yellow_circle: SUGGESTION — L390: `validator_index` is now a dead parameter

**Issue:**
The fix removed `GetKappa()[validator_index].Ed25519`, which was the only use of
`validator_index` inside `BuildJudgements`. After the change the parameter is no longer
referenced anywhere in the function body. Go does not error on unused parameters, but
keeping it here is misleading — it suggests the signature is still tied to a validator
entry in chain state, which is exactly the behaviour this PR moved away from.

**Current code (L386-L391):**
```go
func BuildJudgements(
	tranche types.U8,
	auditReports []types.AuditReport, // (c, w) ∈ aₙ
	hashFunc func(types.ByteSequence) types.OpaqueHash,
	validator_index types.ValidatorIndex,
	validatorPrivKey ed25519.PrivateKey, // κ[v]ᵉ: Ed25519 private key
) []types.AuditReport {
```

**Suggested fix:** Either drop the parameter and update the two call sites (L607, L648)…

```go
func BuildJudgements(
	tranche types.U8,
	auditReports []types.AuditReport, // (c, w) ∈ aₙ
	hashFunc func(types.ByteSequence) types.OpaqueHash,
	validatorPrivKey ed25519.PrivateKey, // κ[v]ᵉ: Ed25519 private key
) []types.AuditReport {
```

…or, if it is kept for API symmetry with `BuildAnnouncement`, add a short comment
explaining why an otherwise-unused parameter is retained.

**Why:** Removes dead state from the public signature and avoids implying a chain-state
lookup that no longer happens.

---

### `internal/auditing/stage2_test.go`

#### :blue_circle: NIT — L52-L62: `TestStage2_GetTranchIndex` effectiveness depends on `TranchePeriod`

**Issue:**
The slot is placed ~120s before "now" to keep `n` away from 0. If `TranchePeriod` (`A`)
is ≥ 120 in some test mode, `n` collapses to 0 — the test still passes but no longer
exercises a meaningful (non-zero) tranche index.

**Current code (L52-L62):**
```go
// Place the block's slot ~120 seconds before "now" so n ≈ 120/A is a
// stable, non-trivial answer (away from 0).
const targetSecondsBack = 120
nowSec := header.GetCurrentTimeInSecond()
slot := types.TimeSlot((nowSec - targetSecondsBack) / uint64(types.SlotPeriod))
```

**Suggested fix:** Derive the offset from `A` (e.g. `3 * types.TranchePeriod`) so the
expected `n` stays well above 0 regardless of test mode.

**Why:** Keeps the test discriminating across `tiny`/`full` (and any future) modes instead
of silently degrading to the trivial `n = 0` case.

YCC3741 (changes requested):
As the above-mentioned comments.

internal/auditing/auditing.go
```diff
@@ -382,12 +385,18 @@ func EvaluateReport(
...
+//
+// Each judgment is signed with the validator's Ed25519 private key. The chain-state
+// entry Kappa[v].Ed25519 is the public key (Ed25519Public, 32B); passing it to
+// ed25519.Sign panics with "bad private key length: 32". The private key must
+// be supplied by the caller (same pattern as BuildAnnouncement).
 func BuildJudgements(
 	tranche types.U8,
 	auditReports []types.AuditReport, // (c, w) ∈ aₙ
 	hashFunc func(types.ByteSequence) types.OpaqueHash,
 	validator_index types.ValidatorIndex,
```
yu2C: I suggest using camel case for `validatorIndex`.

internal/auditing/auditing.go
```diff
@@ -382,12 +385,18 @@ func EvaluateReport(
...
 
-// (17.18) n = {Sκ[v]e (Xe(w) ⌢ H(w)) S (c, w) ∈ an}
+// (GP §17.17, v0.7.2; older revisions §17.18) jn = {S_κ[v]ᵉ ⟨Xe(w) ⌢ H(w)⟩ | (c, w) ∈ aₙ}
+//
+// Each judgment is signed with the validator's Ed25519 private key. The chain-state
+// entry Kappa[v].Ed25519 is the public key (Ed25519Public, 32B); passing it to
+// ed25519.Sign panics with "bad private key length: 32". The private key must
+// be supplied by the caller (same pattern as BuildAnnouncement).
 func BuildJudgements(
 	tranche types.U8,
```
yu2C, line 395: Are we planning to use this argument? If not, we should consider removing it from the signature.

yu2C (commented):
Just some nits.

YCC3741 (approved):

Issue #953, opened 2026-04-27
test(telemetry): e2e against JamTART


HanaYukii:
## Scope

End-to-end integration test against JamTART.

- docker-compose harness next to `cmd/node/`
- Scenarios A-E: happy path / JamTART restart / high load / coverage / 1-hour soak
- Coverage report: JIP-3 events (115) vs JamTART-decoded vs node-emitted, attached to PR description
- Followup issues filed for events JamTART doesn't decode

Parent: #775. Needs #943 foundation plus enough of #958's checkboxes merged that there's real event flow (status + networking + block distribution at minimum).

## Acceptance

- [ ] Scenarios A-E pass
- [ ] Coverage table attached to PR description
- [ ] Nightly contract suite green for 7 days before merge
- [ ] Reviewer signoff

Issue #946, opened 2026-04-27
feat(telemetry): Block Outline type + codec


HanaYukii:
## Scope

Pure type + codec for `Block Outline`, used by events 42 / 43 / 68:

```
u32 size ++ Header Hash ++ u32 tickets ++ u32 preimages ++ u32 preimage total bytes
++ u32 guarantees ++ u32 assurances ++ u32 dispute verdicts
```

No emission, just the type and its encode/decode. Unblocks the block-related checkboxes in #958.

Parent: #775. Independent of #943 and the main loop, can land in parallel.

## Acceptance

- [ ] Encode/Decode roundtrip
- [ ] Golden vector (GP block fixture if available; synthetic otherwise)
- [ ] Reviewer signoff

Issue #775, opened 2025-11-14
JIP-3: Telemetry


yu2C:
https://github.com/polkadot-fellows/JIPs/blob/main/JIP-3.md
Nodes shall expose a CLI option `--telemetry HOST:PORT`. When this option is provided, the node must establish a TCP/IP connection to the specified endpoint.

- Implement a telemetry client that encodes/decodes the common types and event payloads defined in JIP-3.
- Provide helper utilities ( or macros ) so each event handler can easily emit the corresponding telemetry message.
- Test end-to-end against JamTART (see https://github.com/paritytech/jamtart/releases).

HanaYukii:
## JIP-3 Telemetry — Planning

A Go telemetry client implementing [JIP-3](https://github.com/polkadot-fellows/JIPs/blob/main/JIP-3.md): the node opens a TCP connection to a [JamTART](https://github.com/paritytech/jamtart) aggregator and reports observability events as they happen. ~5000 LOC delivered across 4 sub-PRs to a parent integration branch.

### Architecture

```
  caller goroutines (Emit / EmitLazy / EmitFollowup)
       │
       ▼  one at a time (sequencer mutex): assign event ID, queue or mark dropped
  ┌────────────────┐         ┌──────────────────┐
  │ send queue     │         │ dropped events   │
  │ (Go channel)   │         │ (compact ranges) │
  └────────┬───────┘         └────────┬─────────┘
           │                          │
           └────── writer drains ─────┘
                        │  encode + frame  (4-byte LE length prefix)
                        ▼
                    TCP → JamTART
```

`Emit` never blocks the caller. When the queue is full, the lost event IDs are tracked in a "drop range" so the receiver knows exactly which IDs went through and which were dropped. A single writer goroutine drains queue + drop list in order, keeping wire ordering aligned with caller-side ID allocation.

### Wire-correctness invariants

The non-obvious correctness properties the foundation must guarantee before any domain event lands:

- **Producer serialization**: ID allocation + try-send-or-drop happens under one sequencer mutex, so wire order matches event ID order under N concurrent producers.
- **Stale parent rejection**: event IDs pack `(epoch << 48) | seq`; epoch increments on every reconnect; `EmitFollowup` with a parent from a previous epoch returns `InvalidID` so cause-effect chains never cross reconnect boundaries.
[...]

PR #954, opened 2026-04-27
feat: JIP-3(telemetry) TCP client foundation for node observability

### LINKED ISSUE #943: feat(telemetry): foundation TCP client + framing + Node Info
## Scope

Foundation for the JIP-3 telemetry client. One PR (#954, draft).

Wire: TCP client, 4-byte LE length framing, Node Info encoding (no Timestamp / Discriminator header), reconnect with exponential backoff (1s→30s).

Internal: Client API (`Enabled` / `Emit` / `EmitLazy` / `EmitFollowup` / `EmitFollowupLazy` / `Close`); event ID packed `(epoch << 48) | seq` so stale parents from prior connections are rejected; multi-range drop window guarded by sequencer mutex; writer goroutine with panic recover; `InvalidID = ^uint64(0)` sentinel.

CLI: `--telemetry HOST:PORT` flag wired into `cmd/node/main.go`.

Common-type codec: `Option<T>`, `String<N>` (UTF-8 + len cap), `len++[T]`, `bool`, fixed-int LE, `Reason`.

Also lands the contract test framework and `testdata/wire_fixtures/` so domain PRs can drop their golden vectors in.

Parent: #775. Gates everything else.

## Acceptance

- [ ] Codec roundtrip + golden vector tests pass
- [ ] mockTcpServer integration: lifecycle, reconnect, concurrent emit, interleaved drop, bridge invariants, panic recover, Close
- [ ] E2E docker-compose JamTART: Node Info on `/api/ws`, stress overflow exercises Dropped event ID alignment
- [ ] Reviewer signoff

HanaYukii:
> 📐 Project planning, architecture, wire-correctness invariants, and test layers live in the [planning comment on #775](https://github.com/New-JAMneration/JAM-Protocol/issues/775#issuecomment-4324368926).

## What this is

A Go client for [JIP-3 telemetry](https://github.com/polkadot-fellows/JIPs/blob/main/JIP-3.md): the node opens a TCP connection to a [JamTART](https://github.com/paritytech/jamtart) aggregator and reports events (block imported, peer added, etc.) as they happen.

**This PR is just the foundation — no events get reported yet.**

Covers:

- Encoding format (codec primitives + `NodeInfo`)
- TCP connection (4-byte LE framing, dial, exp-backoff reconnect)
- Send queue (buffered channel, drop handling, sequencer mutex)
- Lifecycle (`Close` with force-cancel, panic recover → degraded)
- `--telemetry HOST:PORT` CLI flag

Not covered (lands in later sub-PRs):

- Actual event reporting from node code — block authoring, networking, work-package handling, etc. → #946, #958
- `telemetry.Bridge` helper (no caller for it yet) → #946
- Live JamTART E2E (foundation already proves wire alignment against a mock aggregator) → #953

## Flow

```
  caller goroutines (Emit / EmitLazy / EmitFollowup)
       │
       ▼  one at a time (mutex): give it an ID, then queue or mark as dropped
  ┌────────────────┐         ┌──────────────────┐
  │ send queue     │         │ dropped events   │
  │ (Go channel)   │         │ (compact ranges) │
  └────────┬───────┘         └────────┬─────────┘
           │                          │
           └────── writer drains ─────┘
                        │  encode + frame  (4-byte length prefix per message)
                        ▼
                    TCP → JamTART
```

`Emit` never blocks the caller. If the queue is full, the lost event IDs are tracked in a "drop range" (a window of consecutive lost IDs) so the receiver knows exactly which IDs went through and which were dropped. A single writer goroutine drains the queue and the drop list in order, keeping the wire ordering matching what the caller saw.

## Files

| File | What it does |
|---|---|
| `client.go` | `Client` interface + `InvalidID` sentinel + disabled no-op impl + `New()` factory |
| `codec.go` | JAM codec primitives — LE ints, bool, varlen nat, `len++[u8]`, `String<N>`, `Reason`, `Option<T>` |
| `nodeinfo.go` | `NodeInfo` struct + wire encoder + `NodeFlags` validation |
| `sequencer.go` | Sequencer mutex (atomic ID alloc + drop record), `epoch++` on reconnect, event ID layout helpers |
| `dropranges.go` | Multi-range drop state (disjoint, contiguous-merge) + runtime invariant guard |
| `tcp.go` | `tcpClient` struct + `Emit*` methods + envelope + connect / reconnect loop + `Close` force-cancel |
| `writer.go` | Writer goroutine — lazy dispatch, leading-drop flush, tail-drop ticker, frame encoders, `writeAll` |
| `config.go` | `Config` struct + defaults + validation (negative durations, inverted Reconnect bounds) |
| `cmd/node/main.go` | `--telemetry HOST:PORT` CLI flag wired to `telemetry.New` |
| `*_test.go` | Per-unit tests next to each source file; `stress_test.go` covers concurrent producers / overflow / close drain |

## Open TODOs (intentional, marked in code)

- Canonical `JAMParameters` encoder (waiting on GP / JIP-3 / JamTART reference).
- JAM Common Era epoch for timestamps (currently using Unix μs).

## Branch

Target `feat/jip3-telemetry` (parent integration branch). Parent merges to `main` in one shot once all sub-PRs land.

## Tested

```bash
go test -mod=mod -race -count=1 ./internal/telemetry/   # all pass
```

Two rounds of AI review findings were addressed before the squash. The 8 issues across both rounds (drop flush race, followup epoch race, Close force, short write, NodeFlags, writeNodeInfo+Close blind spot, reader connWG, epoch wrap, config validation, dropState invariant guard, builder purity contract) and per-finding rationale live in the PR comments.

Closes #943 via parent merge.


HanaYukii:
Round 2: thanks to both reviewers — the writeNodeInfo finding (independently flagged by both) was a real blind spot in the round-1 force-close fix; nice catch. 7 fixes pushed in `8c5c8b54`.

**Important — writeNodeInfo + Close blind spot** *(both reviewers)*. `currentConn` was assigned after `writeNodeInfo` returned, so a peer that hangs on the very first `Write` (half-open TCP, kernel retransmit pending up to ~15min) would make Close's force path see `nil` and skip force-close. Fix: assign `currentConn` immediately after a successful dial; clear it in the writeNodeInfo-failure branch. Test: `TestClose_ForceClosesDuringNodeInfoBlock` pins writeNodeInfo with a blocking-Write conn and asserts Close returns within bound.

**Important — reader goroutine not in connWG** *(R2)*. Real contract violation; across reconnects, multiple readers could briefly coexist past Close. Fix: track in `connWG`. Test: `TestClose_WaitsForAllReadersAcrossReconnects` does 3 reconnect cycles and asserts Close stays bounded.

**Important — 16-bit epoch wrap** *(R1)*. Real, ~18h to reach in a flapping accept-then-close worst case. `bumpEpoch` now returns false on wrap; connectLoop degrades cleanly rather than wrapping to 0 (which would silently validate zero-value parent IDs and alias stale parents). Test: `TestSequencer_BumpEpochRefusesWrap` actually loops 65534 bumps and asserts the next refuses without state mutation.

**Minor — reader busy-loop on `(0, nil)`** *(R2)*. `net.TCPConn` doesn't return this but a layered Reader could. Defense: 10ms sleep + peerClosed escape hatch.

**Minor — negative-duration config validation** *(R1)*. Negative `TailDropInterval` panics `time.NewTicker`; negative reconnect intervals hot-loop. Added `Config.validate()` — rejects negative durations and inverted `Reconnect` bounds. Test: `TestNew_RejectsInvalidDurations` covers all 5 paths.

**Minor — `dropState.record` runtime invariant** *(R2)*. Doc said "caller-enforced strictly-increasing ids" but no runtime check. Cheap O(1) panic on violation so a future refactor breaking the sequencer mutex pairing fails loudly rather than silently misordering wire output. 4 tests covering duplicate / decreasing / inside-range / valid happy paths.

**Minor — `EmitLazy` builder purity contract** *(R2)*. Doc was unclear that builders may be silently skipped (drop / disabled / degraded / drain). Added explicit "MUST be pure encoder; side effects may be silently skipped" line in `Client` interface doc.

**Tightening — stress overflow test** *(R2)*. `BufferSize=4 + 200 emits` was theoretically flake-prone on fast CI. Now `BufferSize=1 + 500 emits` guarantees buffer-full regardless of CI speed.

**Not changed (but acknowledged)**:
- *Degraded mode is sticky* — design call per design-doc invariant 3 (transient panic shouldn't auto-recover; explicit operator action required). Reasonable to revisit in a later PR with `MaxPanics` or self-heal policy.
- *`validateParent` (locking variant) is still used by `sequencer_test.go`* — not dead code, kept for test convenience.

Race-clean, full suite passes:
```
go test -mod=mod -race -count=1 ./internal/telemetry/
```

HanaYukii:
TL;DR for late readers: Close now tracks both the active conn and the reader goroutine; the force path covers all four blocking points — `net.DialContext`, `writeNodeInfo` write, `writeLoop` writes, and the read-side liveness watcher. Combined with the round-1 fixes, that closes the lifecycle-correctness loop for the foundation.

yu2C (approved):
Solid JIP-3 foundation and test coverage; a few small follow-ups:

- **READMERef / links:** Drop a brief note under **`READMERef/`** (how to turn telemetry on, `Close`, packed ID vs wire `seq`) when you get a minute, and swap the **`docs/jip3-telemetry-design.md`** pointers in **`client.go`** / **`cmd/node/main.go`** to that—or add the missing doc—so nobody hits a dead link.

- **`Emit*` docs:** Mention that the returned **`uint64` is `(epoch << 48) | seq`**; JamTART’s per-connection numbering lines up with **`seq`** only.

- **`sequencer`:** Optional guard or comment if **`seq`** could ever exhaust the **48-bit** lane on one long-lived TCP session (marginal).

- **`JAMParameters`:** Hook **`NodeInfo.JAMParameters`** to canonical GP/host encoding when available (#775); empty blob is acceptable for foundation.

PR #957, opened 2026-04-27
test: ComputeAnForValidator threshold + a0 top-10 cap

### LINKED ISSUE #956: test: ComputeAnForValidator stochastic VRF threshold (GP §17.14-17.15)
## Context

`ComputeAnForValidator` (`internal/auditing/auditing.go:280-355`) computes the per-tranche stochastic audit assignment. For each work report, it:

1. Builds context ⟨XU ⌢ Y(Hv) ⌢ H(w) ⌢ n⟩
2. VRF-signs context with the validator's Bandersnatch key
3. Extracts VRF output `sₙ(w)`
4. Compares `sn_w[0] · V / (256 · F) < mₙ` to decide whether this validator is assigned to audit `w` in tranche `n`

`F = types.BiasFactor = 2` per GP §17.16 — Burdges/Cevallos 2024 modeling shows this gives one no-show in a tranche an expected `F = 2` validators picking it up the next round, the cryptoeconomic recovery invariant for audit.

**Currently 0 tests.** A wrong-direction comparison, F=1 instead of F=2, division swapped with multiplication, or threshold inverted would all silently break audit recovery and no existing test would catch it.

## Suggested test plan

Two layers, low cost together:

1. **Threshold logic, mocked VRF output** — extract or directly drive the threshold check with synthetic `sn_w[0]` values around the boundary `256·F·mₙ/V`:
   - Just below threshold → assigned
   - Just above threshold → not assigned
   - Exactly at threshold → check the edge consistently

   If the threshold check is hard to isolate from the VRF call, a small refactor to extract `func isAssignedByThreshold(snWByte0 byte, validatorsCount, biasFactor, noShowCount int) bool` would make this unit-testable and is a wider win for readability.

2. **(Optional) statistical** — fixed seed Bandersnatch VRF, run N=1000 reports through `ComputeAnForValidator`, assert assigned count is within ±10% of `F · mₙ`. Slightly flaky in principle but with a fixed seed should be stable. Skip if Bandersnatch fixture plumbing is too heavy for this issue's scope.

## Out of scope

- §17.16 `eₙ` via Refine (PVM cost instrumentation)
- `SingleNodeAuditingAndPublish` end-to-end (separate followup)

## References

- Audit test plan: #931
- Spec: GP §17.14-17.15, §17.16 (BiasFactor)
- Original Stage 4 plan that included this as a stub before the branch was obsoleted: #931 stage 4 section

HanaYukii:
## What

Two coverage gaps left over after stages 1-3 landed:

| Gap | GP | Function |
|---|:-:|---|
| Stochastic audit assignment threshold formula has 0 tests; F = 2 cryptoeconomic recovery invariant unchecked | §17.15 | `ComputeAnForValidator` |
| Top-10 cap branch never reached (stage 1 test used 3 reports) | §17.5 | `buildInitialAuditAssignmentFromCoreOrder` |

Closes #956.

## Refactor

`ComputeAnForValidator`'s threshold check  `Y(sₙ(w))[0] · V / (256 · F)  <  mₙ`  was inlined and tied to the Bandersnatch VRF call producing `sₙ(w)` (Rust FFI). Extracted to a pure helper `isAssignedByThreshold(snByte0, V, F, mn)` with byte-for-byte identical arithmetic so it can be unit-tested without VRF plumbing.

## Tests (4 functions, 18 subtests)

| Function | Cases | GP |
|---|---|:-:|
| `isAssignedByThreshold` boundary | tiny (V=6) and full (V=341), mn = 1, 2; snByte0 just below / just above integer-truncation boundary; defensive `mn = 0` (10 subtests) | 17.15 |
| `isAssignedByThreshold` exhaustive | walk all 256 byte values, assert assigned count matches analytic ⌈256·F·mn/V⌉ for 5 (V, F, mn) combos | 17.15 |
| `buildInitialAuditAssignmentFromCoreOrder` top-10 cap | 12 non-nil reports in shuffle order, assert `len(a0) == 10` | 17.5 |
| Same, with nils | 14 cores with 3 nil entries scattered in, assert nils skipped while 10-cap counts only non-nil | 17.5 |

## Tested

\`\`\`bash
go test -mod=mod -v -run "TestIsAssignedByThreshold|TestBuildInitialAuditAssignmentFromCoreOrder_" ./internal/auditing/  # 18 subtests pass
go test -mod=mod ./internal/auditing/                                                                                     # full pkg pass
\`\`\`

WSL only.

### DIFF
--- a/internal/auditing/audit_followup_test.go
+++ b/internal/auditing/audit_followup_test.go
@@ -70,8 +70,11 @@ func TestIsAssignedByThreshold_BoundaryCases(t *testing.T) {
 }
 
 // Exhaustive: walk all 256 byte values for fixed (V, F, mn) and assert the
-// total number of "assigned" outcomes matches ⌈256·F·mn/V⌉. This pins the
-// formula's behavior across the full input domain in one go and catches
+// total number of "assigned" outcomes matches min(256, ⌈256·F·mn/V⌉). The
+// 256 cap matters once the threshold reaches the byte domain: e.g. tiny
+// mn=3 gives ⌈256·2·3/6⌉ = 256, so every X qualifies and the count
+// saturates rather than overflowing the 256-value range. This pins the
+// formula's behaviour across the full input domain in one go and catches
 // off-by-one errors at the threshold edge that a sparse boundary test
 // might miss.
 func TestIsAssignedByThreshold_ExhaustiveByteRangeCount(t *testing.T) {
--- a/internal/auditing/audit_followup_test.go
+++ b/internal/auditing/audit_followup_test.go
@@ -1,0 +1,183 @@
+// Followup tests for two coverage gaps reviewers flagged after stages 1-3
+// landed: GP §17.15 ComputeAnForValidator threshold formula, and GP §17.5
+// initial-assignment 10-core cap. See issues #931, #956.
+package auditing
+
+import (
+	"testing"
+
+	"github.com/New-JAMneration/JAM-Protocol/internal/types"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// ===========================================================================
+// GP §17.15 isAssignedByThreshold
+//
+//	Y(sₙ(w))[0] · V / (256 · F)  <  mₙ   ⇒  validator is assigned
+//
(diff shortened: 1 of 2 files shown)


YCC3741 (approved):
NICE

Issue #958, opened 2026-04-27
feat(telemetry): JIP-3 domain events umbrella (#775)


HanaYukii:
## Scope

JIP-3 domain events, 115 across 8 domains. Each checkbox is one PR off `feat/jip3-telemetry`. Foundation API comes from #943, Block Outline from #946.

- [ ] **Status (10-13)** — 4 events. Needs Best Block Changed publisher (no existing EventType) plus publishers for `BlockFinalized` / `BulkSyncCompleted` (EventTypes exist but have 0 callsites). 2-second status loop emits event 10.
- [ ] **Networking (20-28)** — 9 events. QUIC peer / connection lifecycle hooks in `internal/networking/quic/`.
- [ ] **Block authoring/import (40-47)** — 8 events. Event 47 emits Accumulate Cost as zeros per Q2 resolution.
- [ ] **Block distribution (60-68)** — 9 events. Block announcement + request hooks.
- [ ] **Safrole + Preimage (80-84, 190-199)** — 15 events. Ticket VRF flow + `ce128 / ce129`.
- [ ] **Guaranteeing (90-113)** — 24 events, the biggest single domain. Hooks across `ce14{2,3,5,7}` + `internal/work_package/`. Events 95 / 101 emit cost structs as zeros per Q2; PVM cost instrumentation is a separate future prerequisite.
- [ ] **Availability + Bundle (120-153)** — 26 events. Investigate whether the shard distribution layer exists during PR.
- [ ] **Segment recovery (160-178)** — 19 events. `ce140 / ce141`.

Each checkbox needs the node main loop running; coordinate with #774 / networking work touching `cmd/node/main.go`.

Parent: #775. Replaces the original per-domain sub-issues (#944, #945, #947, #948, #949, #950, #951, #952), all closed and consolidated here.

Issue #774, opened 2025-11-14
JIP-2: Node RPC


yu2C:
https://github.com/polkadot-fellows/JIPs/blob/dave/jip-2-fixes/JIP-2.md
Based on JSON-RPC 2.0 format, and the implementation priority is
1. Type definitions ( common types, error codes, descriptors )
2. Methods
    - Block related
    - Service related
    - Statistics related
    - Work Packages related
    - Sync state
3. Chain subscription

PR #961, opened 2026-05-07
feat(telemetry): Block Outline + Cost structs (#946)

### LINKED ISSUE #946: feat(telemetry): Block Outline type + codec

HanaYukii:
> Next sub-PR after #954. See the [planning comment on #775](https://github.com/New-JAMneration/JAM-Protocol/issues/775#issuecomment-4324368926) for the wider plan.

## What this is

Pure types + codec for the JIP-3 Block Outline and the four PVM Cost structs. No emission, no callsite hooks — just the types so the upcoming domain PRs (#958) can build payloads cleanly.

Per #775 Q2, Cost structs ship zero-filled until PVM cost instrumentation lands (a separate prerequisite). A zero-value struct encodes to all-zero bytes so emitters can pass zero without placeholder logic.

## What's in

| Type | Size | Used by |
|---|---|---|
| `BlockOutline` | 60 B | events 42, 43, 68 |
| `ExecCost` | 16 B | nested in the Cost types below |
| `IsAuthorizedCost` | 40 B | event 95 |
| `RefineCost` | 104 B | event 101 |
| `AccumulateCost` | 140 B | event 47 |

Each type has `Encode() []byte` and a matching `Decoder.ReadXxxCost`. Same shape as the existing `NodeInfo` codec.

## Tested

```bash
go test -mod=mod -race -count=1 ./internal/telemetry/   # all pass
```

21 new tests across `blockoutline_test.go` + `cost_test.go`: encoded size matches spec, `Encode → Read` roundtrip, decoder consumes exactly the encoded bytes, every prefix-truncation errors (no silent zeroing), zero-value encodes to all-zero bytes (Q2 contract), and byte-for-byte golden vectors for all 5 types (the `AccumulateCost` one pins `TotalTransferGas`'s position between `Transfer` and `Other`).

## Open TODOs (intentional)

Golden vectors are self-consistent (generated from the encoder we commit). Replace with externally-verified fixtures once GP / JIP-3 / JamTART reference data is available.

## Branch

Targets `main` directly — same merge style as #954 (the parent integration branch wasn't used in the end). Independent of #774 / Q1.

Closes #946.

### DIFF
--- a/internal/telemetry/cost_test.go
+++ b/internal/telemetry/cost_test.go
@@ -1,0 +1,345 @@
+package telemetry
+
+import (
+	"bytes"
+	"encoding/hex"
+	"testing"
+)
+
+// makeSampleExecCost returns an ExecCost with deterministic content.
+func makeSampleExecCost(seed uint64) ExecCost {
+	return ExecCost{
+		GasUsed:      seed,
+		ElapsedNanos: seed*2 + 1,
+	}
+}
+
+func TestExecCost_EncodedSize(t *testing.T) {
+	if got := len(makeSampleExecCost(1).Encode()); got != execCostEncodedSize {
+		t.Errorf("ExecCost size = %d, want %d", got, execCostEncodedSize)
+	}
+}
+
+func TestExecCost_Roundtrip(t *testing.T) {
+	want := makeSampleExecCost(0xDEADBEEF)
+	got, err := NewDecoder(want.Encode()).ReadExecCost()
+	if err != nil {
+		t.Fatalf("ReadExecCost: %v", err)
+	}
+	if got != want {
+		t.Errorf("roundtrip mismatch:\n got %+v\nwant %+v", got, want)
+	}
+}
+
+// Zero-fill: AccumulateCost / IsAuthorizedCost / RefineCost are emitted
+// zero-filled per #775 Q2 until PVM cost instrumentation lands. Encode
+// must produce all-zero bytes for a zero-value struct so callers can
+// rely on it.
+func TestExecCost_ZeroEncodesToZeroes(t *testing.T) {
+	enc := ExecCost{}.Encode()
+	if len(enc) != execCostEncodedSize {
+		t.Fatalf("encoded size = %d, want %d", len(enc), execCostEncodedSize)
+	}
+	for i, b := range enc {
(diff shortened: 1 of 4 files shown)


internal/telemetry/cost_test.go
```diff
@@ -0,0 +1,333 @@
...
+	if got != want {
+		t.Errorf("roundtrip mismatch:\n got %+v\nwant %+v", got, want)
+	}
+}
+
+// Zero-fill: AccumulateCost / IsAuthorizedCost / RefineCost are emitted
+// zero-filled per #775 Q2 until PVM cost instrumentation lands. Encode
+// must produce all-zero bytes for a zero-value struct so callers can
+// rely on it.
+func TestExecCost_ZeroEncodesToZeroes(t *testing.T) {
```
yu2C, line 38: Suggested test hardening (zero-value encode)

The Test*ZeroEncodesToZeroes helpers in` cost_test.go` only iterate `for i, b := range enc`. If `Encode()` ever regressed to returning an empty slice, the loop would not run and the test would still pass, even though the wire size contract is violated.

`TestBlockOutline_ZeroEncodesToZeroes` already guards this with an explicit length check. Consider mirroring that pattern for each cost type, e.g. right after `enc := …Encode()`:
```
if len(enc) != execCostEncodedSize {
	t.Fatalf("encoded size = %d, want %d", len(enc), execCostEncodedSize)
}
```
(and the same for `isAuthorizedCostEncodedSize`, `refineCostEncodedSize`, `accumulateCostEncodedSize` in the respective zero tests.)

yu2C (approved):

PR #970, opened 2026-05-15
feat(telemetry): bridge helpers + contract test framework

HanaYukii:
> Follow-up to #954 and #961. This is part of the JIP-3 telemetry plan in #775.

## Summary

This PR adds helper functions for connecting `internal/networking/quic.EventBus` handlers to the telemetry `Client` API.

No real node events are subscribed in this PR. The domain event PRs (#958) will use these helpers later.

## What Changed

- `Bridge`: encode the event inside the EventBus handler, then call `tel.Emit`.
- `BridgeLazy`: copy the needed event data now, then encode later in the telemetry writer.
- `BridgeFollowup`: get `(parentID, payload)` from each event, then call `tel.EmitFollowup`.
- `BridgeContract`: a small registry for contract tests. Domain PRs can register their bridge and a sample event.
- `quic.Event` is now an alias for `any`, so bridge handlers can be passed directly to `quic.EventBus.Subscribe`.
- A compile-time test checks that all bridge helpers return a valid `quic.Handler`.

## Behavior

Bridge handlers are best-effort telemetry hooks:

- They return `nil` to the EventBus.
- They do not do network I/O in the EventBus path.
- They recover synchronous panics from `encoder`, `snapshot`, or `extract` and drop that telemetry event.
- `BridgeLazy` runs `snapshot` immediately. The snapshot must copy mutable data from the event. Do not return the original event pointer.
- `BridgeLazy` runs `encode` later in the telemetry writer goroutine. If that encode function panics, the existing writer recovery marks telemetry as degraded.

## Not In This PR

- No `eb.Subscribe(...)` callsites yet.
- No domain-specific telemetry events yet.
- No nightly CI runner for the contract registry yet.

## Files

| File | Purpose |
|---|---|
| `internal/telemetry/bridge.go` | Bridge helper APIs |
| `internal/telemetry/contract.go` | Contract registry |
| `internal/networking/quic/event_bus.go` | Makes `Event` an alias for `any` for handler compatibility |
| `internal/telemetry/*_test.go` | Contract, compatibility, panic, lazy snapshot, and race tests |

## Tested

```bash
go test -mod=mod -race -count=1 ./internal/telemetry/ ./internal/networking/quic/
```

## Branch

Targets `main` directly. Independent of #774 / Q1 and #946.

Closes the bridge-helper part of #943. Unblocks #958.

### DIFF
--- a/internal/telemetry/bridge_test.go
+++ b/internal/telemetry/bridge_test.go
@@ -1,0 +1,357 @@
+package telemetry
+
+import (
+	"context"
+	"sync"
+	"sync/atomic"
+	"testing"
+	"time"
+)
+
+// ---------------------------------------------------------------------------
+// Fake Client for tracking bridge behaviour. Not thread-safe across
+// inspection / call (tests inspect only after the handler returns).
+// ---------------------------------------------------------------------------
+
+type fakeClient struct {
+	enabled atomic.Bool
+
+	emitCount     atomic.Int32
+	lazyCount     atomic.Int32
+	followupCount atomic.Int32
+
+	// Last-seen fields are guarded by mu so the concurrent-invocation
+	// test (and the race detector) is happy. Single-threaded tests can
+	// read them after the handler returns without locking, but the
+	// helpers below lock anyway to keep the API simple.
+	mu               sync.Mutex
+	emitLastDisc     uint8
+	emitLastPayload  []byte
+	lazyLastDisc     uint8
+	lazyLastBuilder  func() []byte
+	followupLastDisc uint8
+	followupParent   uint64
+	followupPayload  []byte
+}
+
+func newFakeClient(enabled bool) *fakeClient {
+	c := &fakeClient{}
+	c.enabled.Store(enabled)
+	return c
+}
+
+func (c *fakeClient) Enabled() bool { return c.enabled.Load() }
+
+func (c *fakeClient) Emit(disc uint8, payload []byte) uint64 {
+	c.emitCount.Add(1)
+	c.mu.Lock()
+	c.emitLastDisc = disc
+	c.emitLastPayload = append([]byte(nil), payload...)
+	c.mu.Unlock()
+	return uint64(c.emitCount.Load())
+}
+
+func (c *fakeClient) EmitLazy(disc uint8, builder func() []byte) uint64 {
+	c.lazyCount.Add(1)
+	c.mu.Lock()
+	c.lazyLastDisc = disc
+	c.lazyLastBuilder = builder
+	c.mu.Unlock()
+	return uint64(c.lazyCount.Load())
+}
+
+func (c *fakeClient) EmitFollowup(disc uint8, parentID uint64, payload []byte) uint64 {
+	c.followupCount.Add(1)
+	c.mu.Lock()
+	c.followupLastDisc = disc
+	c.followupParent = parentID
+	c.followupPayload = append([]byte(nil), payload...)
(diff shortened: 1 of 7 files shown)


internal/telemetry/bridge_test.go
```diff
@@ -0,0 +1,339 @@
...
+	h := BridgeFollowup(fc, 11, func(ev any) (uint64, []byte) {
+		called = true
+		return 1, nil
+	})
+	_ = h(context.Background(), 0)
+	if called {
+		t.Errorf("extract ran on disabled client")
+	}
+}
+
```
yu2C, line 309: **SUGGESTION**

`Bridge` and `BridgeLazy` already have tests that assert the handler returns `nil` and does not enqueue telemetry when the user closure panics (`TestBridge_ReturnsNilOnEncoderPanic`, `TestBridgeLazy_ReturnsNilOnSnapshotPanic`). `BridgeFollowup` uses the same `defer recoverBridgePanic(disc)` pattern around `extract(ev)`, but there is no matching test that panics inside `extract`.

Please add something like `TestBridgeFollowup_ReturnsNilOnExtractPanic`: enabled fake client, `extract` that panics, assert `h(ctx, ev) == nil` and `followupCount == 0` (no follow-up emit after panic). That keeps invariant coverage symmetric across the three bridge constructors.

internal/telemetry/contract_test.go
```diff
@@ -0,0 +1,112 @@
...
+	Register(BridgeContract{
+		Name:    "synthetic.eager",
+		Disc:    1,
+		Handler: Bridge(fc, 1, func(ev any) []byte { return []byte{0x01} }),
+		Sample:  func() any { return 0 },
+	})
+	Register(BridgeContract{
+		Name: "synthetic.lazy",
+		Disc: 2,
+		Handler: BridgeLazy(fc, 2,
```
yu2C, line 92: **SUGGESTION**

The `BridgeLazy` synthetic registration uses `func(ev any) any { return ev }` as `snapshot`. In `bridge.go`, the `BridgeLazy` documentation explicitly calls out returning `ev` (or a struct that still aliases mutable pointer data) as an anti-pattern because deferred `encode` can race with the publisher mutating the event.

Here `Sample` returns a plain `int`, so there is no pointer aliasing in practice, but this snippet is easy to copy into real domain code where `ev` is a pointer type. Consider making the snapshot copy by value explicitly (e.g. `func(ev any) any { v, _ := ev.(int); return v }`) and add a short comment that `return ev` is only safe for value-only samples so future readers do not generalize it to pointer events.

internal/telemetry/bridge_test.go
HanaYukii: Added in `e8c1f2f6` as `TestBridgeFollowup_ReturnsNilOnExtractPanic` — mirrors the encoder/snapshot panic tests: enabled fake client, extract that panics, assert handler returns nil and `followupCount == 0`.

internal/telemetry/contract_test.go
HanaYukii: Good catch — that snippet was exactly the anti-pattern doc'd in `bridge.go`. Fixed in `e8c1f2f6`: snapshot now does `v, _ := ev.(int); return v` (explicit value copy), plus a comment that `return ev` is only safe when `Sample` returns a value type and real domain bridges with pointer events MUST copy fields out per the bad/good examples in `bridge.go`.

yu2C (approved):

PR #978, opened 2026-05-22
[WIP] refactor(telemetry): lift cost types to internal/pvmcost (#974 Phase 0)

HanaYukii:
Phase 0 of #974. Wire format byte-identical, no runtime behaviour change.

Moves the 4 JIP-3 cost structs (`ExecCost` / `IsAuthorizedCost` / `RefineCost` / `AccumulateCost`) out of `internal/telemetry` into a new leaf package `internal/pvmcost`:

```
PVM ──▶ pvmcost ◀── telemetry   (wire encoding stays in telemetry)
```

PVM has to produce these values but must not import telemetry (TCP / framing / IO). The neutral leaf package fixes the import direction.

**API change (breaking, but no callers yet):** `(c ExecCost).Encode()` → `telemetry.EncodeExecCost(c)`, and `telemetry.ExecCost` is gone — use `pvmcost.ExecCost`. No aliases kept: emission isn't built yet, the only caller was the test file (updated here), so aliases would be dead shims.

**CI guard:** `consensus_boundary_test.go` reflect-walks Block / WorkReport / … and fails if any consensus type ever transitively embeds a pvmcost type — cost is observability-only, never consensus state.

Golden vectors + roundtrip tests pass unchanged. Not included: new events, PVM instrumentation, emission.

### DIFF
--- a/PVM/is_authorized_invocation.go
+++ b/PVM/is_authorized_invocation.go
@@ -1,6 +1,7 @@
 package PVM
 
 import (
+	"github.com/New-JAMneration/JAM-Protocol/internal/pvmcost"
 	"github.com/New-JAMneration/JAM-Protocol/internal/types"
 )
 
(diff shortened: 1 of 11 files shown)


HanaYukii:
Updated the PR description to match the code: this is **not** a pure move — cost encoding changed from `(c X).Encode()` methods to free `telemetry.EncodeXCost(c)` functions, and no type aliases are kept (`telemetry.ExecCost` → `pvmcost.ExecCost`).

Went with **option (b)** — declared the breaking API change rather than adding aliases — because there are no production callers yet (emission / #974 Phase 1 isn't built; only the test file referenced these types, and it's updated here). Aliases would just be dead shims. Wire format stays byte-identical; golden + roundtrip tests pass.

Thanks for catching the description ↔ code mismatch.

PR #990, opened 2026-05-25
fix: prune persistent storage in fuzz mode to prevent disc exhaustion

YCC3741:
Closes #991

## Summary

Fixes disk exhaustion during long fuzz sessions (361K+ steps) by pruning persistent storage when `JAM_FUZZ=1`.

**Root cause:** Every `ImportBlock` writes full blocks (`b:<hash>`) and header mappings (`ht:<hash>`) to Pebble, but never deletes them. State data (`sd:<root>`) was only evicted from memory, not disk. After 361K steps this exhausts container disk space.

**Fix:** When `JAM_FUZZ=1`, evict old data from persistent storage alongside memory eviction, triggered by AncestryCache trim (MaxLookupAge threshold). Normal mode behavior is unchanged.

### Changes

- `internal/store/block.go`: Add `DeleteBlockByHash`
- `internal/blockchain/ancestry.go`: `AppendAncestry` returns evicted items
- `internal/blockchain/chain_state.go`: New `evictOldData` handles disk cleanup; `isFuzzMode()` reads `JAM_FUZZ` env var

### Disk behavior (measured with minifuzz storage suite)

| Round | Steps | Disk usage |
|-------|-------|-----------|
| 1 | ~51 | 21.2 MB |
| 2 | ~102 | 28.9 MB |
| 3 | ~153 | 29.8 MB |
| 4-9 | ~204-459 | ~29.8 MB (stable) |
| 10 | ~510 | 27.1 MB (compaction) |

Disk stabilizes after exceeding MaxLookupAge and no longer grows unboundedly.

### Test results

- Minifuzz (4 suites): PASS
- Picofuzz (4 suites): PASS
- jam-conformance traces: ALL PASSED

## Test plan

- [x] Minifuzz 4 suites pass
- [x] Picofuzz 4 suites pass (no MISMATCH)
- [x] jam-conformance traces all pass (no FAILED)
- [ ] Long fuzz session (>1000 steps) confirms disk stays bounded

### DIFF
--- a/internal/blockchain/chain_state.go
+++ b/internal/blockchain/chain_state.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"os"
 	"sort"
 	"strings"
 	"sync"
@@ -52,10 +53,11 @@ type ChainState struct {
 
 	// cache for leaf level merklization
 	keyLevelCache *KeyLevelCache
+}
 
-	// ring buffer tracking recent state roots stored in repo (memory),
-	// used to evict old entries and bound memory usage.
-	recentStateRoots []types.StateRoot
+func isFuzzMode() bool {
+	v := strings.TrimSpace(os.Getenv("JAM_FUZZ"))
+	return v != "" && v != "0" && v != "false" && v != "no" && v != "off"
 }
 
 func getPersistentDatabase() database.Database {
@@ -413,14 +415,14 @@ func (cs *ChainState) StateCommit() {
 
 		existingAncestry := cs.GetAncestry()
 		if len(existingAncestry) > 0 {
-			// Add to ancestry (avoid duplicating the latest header if it'cs already the last item)
 			currentItem := types.AncestryItem{
 				Slot:       latestBlock.Header.Slot,
 				HeaderHash: blockHeaderHash,
 			}
 			last := existingAncestry[len(existingAncestry)-1]
 			if last.Slot != currentItem.Slot || last.HeaderHash != currentItem.HeaderHash {
-				cs.AppendAncestry(types.Ancestry{currentItem})
+				evicted := cs.AppendAncestry(types.Ancestry{currentItem})
+				cs.evictOldData(evicted)
 			} else {
 				logger.Debugf("StateCommit: latest header already in ancestry (slot=%d, hash=0x%x), skipping append", currentItem.Slot, currentItem.HeaderHash[:8])
 			}
@@ -461,14 +463,6 @@ func (cs *ChainState) StateCommitWithPreComputedState(
 		logger.Warnf("StateCommitWithPreComputedState: failed to store state data to disk: %v", err)
 	}
 
-	// Evict oldest state data from memory when exceeding MaxLookupAge entries.
-	cs.recentStateRoots = append(cs.recentStateRoots, stateRoot)
-	if len(cs.recentStateRoots) > types.MaxLookupAge {
-		evict := cs.recentStateRoots[0]
-		cs.recentStateRoots = cs.recentStateRoots[1:]
-		cs.repo.DeleteStateData(cs.repo.Database(), evict)
-	}
-
 	latestBlock := cs.GetLatestBlock()
 
 	// Persist block mapping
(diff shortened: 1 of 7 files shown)


internal/blockchain/chain_state.go
```diff
@@ -413,14 +415,14 @@ func (cs *ChainState) StateCommit() {
 
 		existingAncestry := cs.GetAncestry()
 		if len(existingAncestry) > 0 {
```
yu2C, line 421: ### [Suggestion] Handle `len(ancestry) == 0` so prune works without `feature-ancestry`

**Context**

- Long-running **serve** / GP fuzz often enables `feature-ancestry` (`SetState` ships ancestors). With this PR, once `len(existingAncestry) > 0`, `StateCommit` → `AppendAncestry` → `evictOldData` can cap persistent storage. That matches the disk exhaustion we saw in long fuzz (e.g. hundreds of thousands of `ImportBlock`s).
- **Conformance / minifuzz / picofuzz** usually run with ancestry **disabled** or an empty `SetState` ancestry. Sessions are short (~102 pairs), so disk growth is not obvious—but the code path still writes persistent state/block every `ImportBlock` while **never** calling `evictOldData` because of `if len(existingAncestry) > 0`.

So: the production issue is tied to **having** ancestry + long runs; the **gap** is the **no-ancestry** path, which remains unbounded if someone runs a long session without `feature-ancestry`.

**Proposed fix (recommended)**

In `StateCommit` (and mirror in `StateCommitWithPreComputedState`), **seed ancestry when empty** instead of skipping the block:

1. Build `currentItem` from the latest committed header.
2. If `len(existingAncestry) == 0`, `AppendAncestry({currentItem})` and `evictOldData(evicted)` (no-op until length exceeds `MaxLookupAge` / fuzz retention cap).
3. Else keep the existing de-dupe + append + evict logic.

This does **not** require GP to send ancestry in `SetState`, fixes prune for our current tests, and still helps serve when ancestry is enabled.

**Optional follow-up**

In `FuzzServiceStub.SetState`, if `len(ancestry) == 0`, synthesize a one-item ancestry from the init header before `AppendAncestry`, and call `evictOldData` after bulk append (for long ancestry lists after `ResetInstance`).

**Not recommended as the only fix**

Requiring `feature-ancestry` in all tests—spec allows it to be off, and disk prune should not depend on GP optional features.

internal/blockchain/chain_state.go
```diff
@@ -413,14 +415,14 @@ func (cs *ChainState) StateCommit() {
...
-			// Add to ancestry (avoid duplicating the latest header if it'cs already the last item)
 			currentItem := types.AncestryItem{
 				Slot:       latestBlock.Header.Slot,
 				HeaderHash: blockHeaderHash,
 			}
 			last := existingAncestry[len(existingAncestry)-1]
 			if last.Slot != currentItem.Slot || last.HeaderHash != currentItem.HeaderHash {
-				cs.AppendAncestry(types.Ancestry{currentItem})
+				evicted := cs.AppendAncestry(types.Ancestry{currentItem})
```
yu2C: ### [Suggestion] Full spec still allows very large persistent retention

Pruning is capped by `types.MaxLookupAge` (L):
- tiny: L = 24
- full (`SetFullMode`): L = **14400**

Even with working prune, full fuzz can retain ~**14400 × (state + block)** on disk; long runs may still fill the disk. Tiny (~24 entries) is much smaller.

**Suggestion:** Add a **fuzz-only** retention constant, separate from protocol L, e.g.:

```go
const FuzzPersistentRetainBlocks = 24 // when JAM_FUZZ is enabled

File: internal/blockchain/chain_state.go [hunk 3]
```diff
@@ -52,10 +53,11 @@ type ChainState struct {
 
 	// cache for leaf level merklization
 	keyLevelCache *KeyLevelCache
+}
 
-	// ring buffer tracking recent state roots stored in repo (memory),
-	// used to evict old entries and bound memory usage.
-	recentStateRoots []types.StateRoot
+func isFuzzMode() bool {
```
yu2C: ### Comment 3 — `internal/blockchain/chain_state.go` (`isFuzzMode`)

**File:** `internal/blockchain/chain_state.go`  
**Around:** `isFuzzMode()`

```markdown
### [Nit] Duplicate `JAM_FUZZ` detection vs `cmd/fuzz`

`isFuzzMode()` duplicates `jamFuzzEnvEnabled()` in `cmd/fuzz/main.go` (empty / `0` / `false` / `no` / `off`).

**Follow-up:** Extract shared helper (e.g. `internal/jamfuzz.Enabled()`); `blockchain` must not import `cmd/fuzz`.

yu2C (changes requested):
### [Suggestion] Call `evictOldData` after bulk `AppendAncestry` in `SetState`

`ResetInstance()` clears in-memory state but not persistent DB. If `len(ancestry) > MaxLookupAge`, `AppendAncestry` trims the cache but does not delete evicted headers from persistent storage.

Suggest: `evicted := cs.AppendAncestry(ancestry); cs.evictOldData(evicted)`.

yu2C (approved):
LGTM

PR #994, opened 2026-05-25
refactor: predecode PVM blocks and instructions

TwEricShen:
Resolves #995

Add block and instruction structure to decode instructions in advance.
Execute the instructions using the decoded data.


PVM/block_info.go
```diff
@@ -0,0 +1,222 @@
...
+				PC:      pc,
+				Opcode:  op,
+				SkipLen: uint8(skipLen),
+			})
+			p.InstrMap[pc] = idx
+
+			// v0.7.2 only — operands are decoded here so the
+			// interpreter can read them from InstrMeta. In v0.8.0 the JIT will be
+			// the sole consumer and decodeOperands should be removed from this loop.
+			decodeOperands(&p.Instrs[idx], idata, bitmask)
```
yu2C, line 195: add error handling?

PVM/instructions_instrmeta.go
```diff
@@ -0,0 +1,2065 @@
...
+	case 87:
+		branchCondition = int64(interp.Registers[rA]) < int64(vX)
+	case 88:
+		branchCondition = int64(interp.Registers[rA]) <= int64(vX)
+	case 89:
+		branchCondition = int64(interp.Registers[rA]) >= int64(vX)
+	case 90:
+		branchCondition = int64(interp.Registers[rA]) > int64(vX)
+	default:
+		pvmLogger.Fatalf("instImmediateBranch is supposed to be called with opcode in [80, 90]")
```
yu2C: **Suggestion (do not kill the whole process)**

`instImmediateBranchMeta` / `instBranchMeta` use `pvmLogger.Fatalf(...)` on the `default` branch (`instructions_instrmeta.go`, and the same in `instructions.go`).

This is **not** gated on the `pvm` logger being enabled. In `logger/logger.go`, `FATAL` always runs and calls `os.Exit(1)` regardless of `enabled` or log level.

So if execution ever reaches that `default` (table drift, bad opcode routing, etc.), the **entire node/test process exits**, instead of returning PVM `ExitPanic` for the invocation.

**Recommendation:** Replace `Fatalf` with `Errorf` + `return ExitPanic, instr.PC` (match other handlers). Keeps conformance/STF behavior as “service panic” rather than aborting the Go runtime.

PVM/opcode_info.go
```diff
@@ -0,0 +1,220 @@
...
+	229: {"min", InstrCatThreeReg, false},
+	230: {"min_u", InstrCatThreeReg, false},
+}
+
+func IsValidOpcode(op byte) bool {
+	return opcodeInfoTable[op].Category != InstrCatInvalid
+}
+
+// IsBlockTerminator reports whether op ends a basic block.
+func IsBlockTerminator(op byte) bool {
```
yu2C, line 207: remove duplicated function in `program_code/isBasicBlockTerminationInstruction ` ?

PVM/instructions_instrmeta.go
```diff
@@ -0,0 +1,2065 @@
...
+		branchCondition = int64(interp.Registers[rA]) < int64(interp.Registers[rB])
+		op = "<(signed)"
+	case 174:
+		branchCondition = interp.Registers[rA] >= interp.Registers[rB]
+		op = ">="
+	case 175:
+		branchCondition = int64(interp.Registers[rA]) >= int64(interp.Registers[rB])
+		op = ">=(signed)"
+	default:
+		pvmLogger.Fatalf("instBranch is supposed to be called with opcode in [170, 175]")
```
yu2C: ditto.

PVM/invocation.go
```diff
@@ -72,6 +73,79 @@ func (interp *Interpreter) SingleStepStateTransition(pc ProgramCounter) (ExitRea
...
+		if block := interp.Program.LookupBlock(pc); block != nil {
+			startIdx = block.InstrStart
+			endIdx = block.InstrEnd
+		} else if idx, ok := interp.Program.InstrMap[pc]; ok {
+			startIdx = idx
+			for endIdx = idx; endIdx < len(interp.Program.Instrs); endIdx++ {
+				if IsBlockTerminator(interp.Program.Instrs[endIdx].Opcode) {
+					endIdx++
+					break
+				}
```
yu2C, line 98: what if no terminator met? it'll keep scanning and cross the block.
suggest add `exitpanic` when no terminator found.

PVM/block_info.go
```diff
@@ -0,0 +1,222 @@
...
+
+// InstrCount returns the number of instructions in this block.
+func (b *BlockMeta) InstrCount() int {
+	return b.InstrEnd - b.InstrStart
+}
+
+// decodeOperands populates the Dst, Src, Imm fields of an InstrMeta
+// by dispatching on the instruction's category and calling the existing
+// decode functions from decode.go. Called once per instruction at deblob time.
+func decodeOperands(instr *InstrMeta, idata ProgramCode, bitmask Bitmask) {
```
yu2C, line 35: **Suggestion (correctness / avoid Go runtime panic)**

`InstrMeta` uses `0xFF` as a sentinel for unused `Dst` / `Src` operands (`block_info.go`). PVM only has registers `0..12` (`Registers [13]uint64`).

In `decodeOperands`, many branches only update fields when `decode*()` succeeds (`err == nil`). On failure, `Dst` / `Src` can remain `0xFF`, but several `*Meta` handlers still index registers directly, e.g.:

- `instMoveRegMeta`: `interp.Registers[instr.Dst] = interp.Registers[instr.Src[0]]`
- similar patterns in `instSbrkMeta`, `instBranchMeta`, load/store `*Meta`, etc.

In Go, `Registers[0xFF]` is `Registers[255]` → **index out of range** (process crash), not PVM `ExitPanic`.

**Recommendation:** Fail fast at deblob time: if any operand decode fails in `decodeOperands`, return `ExitPanic` from `preDecodeBlocks()` (same as invalid opcode). Optionally assert `Dst/Src[i] <= 12` before execution in debug builds.

Valid GP programs should not hit this; the risk is mainly malformed blobs or silent partial decode (e.g. short blob for `load_imm_64` leaves wrong `Imm` but may still set `Dst`).

yu2C (changes requested):

PVM/block_info.go
TwEricShen: I agree with the OOR safety issue. I will do the check in the decode function, if the instruction is invalid, will replace the it with instruction `trap`, which will return `panic` in runtime.

But, the **Recommendation** of fail fast at deblob time will cause different result of PVM. 
The current result return panic at runtime, if an instruction return panic, it might have done some other instructions and charged some gas.
If we do fail fast will cause no gas charging.

PVM/invocation.go
TwEricShen: Basically, correctly pre-decoded programs will not cross the block. Since, we separate the block through the terminator. It will always hit the current block's terminator.
The only failure mode is broken metadata, which would cause wrong execution scope.This indeed needs some defensive check to avoid no terminator found.

yu2C (approved):
LGTM.
but some fmt issue

PR #1000, opened 2026-05-28
perf(types): defer fmt.Sprintf in SCALE codec debug logs

### LINKED ISSUE #999: perf(types): avoid eager fmt.Sprintf in disabled SCALE codec debug logs
## Summary

In the SCALE codec, the debug helper `cLog` is called as
`cLog(color, fmt.Sprintf("...%x", data))` at ~120 sites across
`internal/types/{decode,decoder,encode,encoder}.go`.

Because Go evaluates arguments before the call, the `fmt.Sprintf` (which
hex-encodes whole byte slices and allocates a string) runs on **every**
encode/decode of **every** field — even though `debugMode` is `false` in
production, so `cLog` discards the result immediately.

This is pure wasted work on a very hot path: `StateEncoder` serializes the
entire state twice per block import (prior + posterior state roots), plus
every block decode and DB persist goes through the codec.

## Evidence (pprof, fuzz import path)

Profiling the real import path (`internal/fuzz.ImportBlock`) over current
0.7.2 conformance / picofuzz traces, steady state:

- `fmt.Sprintf` ~10% cumulative, under `types.(*Encoder).Encode` / `ByteSequence.Decode`
- `fmt.(*pp).fmtBytes` ~7%
- On encode-heavy short paths the same pattern reached ~38%.

## Proposed fix

Make `cLog` variadic and move the formatting inside the `debugMode` guard
(lazy), then drop the `fmt.Sprintf` wrapper at the call sites. No
behavioural change when `debugMode` is enabled.

## Scope

`internal/types` codec only. The same `var debugMode = false` + eager
`fmt.Sprintf` pattern exists in other packages (`jamtests/*`,
`internal/utilities/merklization`) and can be addressed as follow-ups.

YCC3741:
## Summary

Removes a wasted-work pattern in the SCALE codec (and the merklization /
jamtests code that shares it), aligns the test-vector submodules with the
GP 0.7.2 implementation, and fixes a test that pointed at stale 0.7.1 data.

| # | Commit | What |
|---|--------|------|
| 1 | `perf(types)` | Make `cLog` variadic and defer `fmt.Sprintf` inside the `debugMode` guard; unwrap call sites in `internal/types/{decode,decoder,encode,encoder}.go` (~120 sites) |
| 2 | `chore(test-data)` | Bump `jam-test-vectors` submodule v0.7.0 -> v0.7.2 |
| 3 | `perf` | Same lazy-`cLog` change in `internal/utilities/merklization` (production) and the `jamtests/*` runners (16 sites) |
| 4 | `chore(test-data)` | Bump `jam-conformance` submodule to 264a4f3 |
| 5 | `fix(test)` | Point `TestFuzzReportsTrace` at `fuzz-reports/0.7.2/traces` (was hardcoded 0.7.1, incompatible with the 0.7.2 impl) |

Closes #999

## Why the perf change matters

`cLog` no-ops when `debugMode` is `false` (production), but callers passed
`fmt.Sprintf("...%x", data)` as an argument, so Go evaluated the (discarded)
hex formatting on every encode/decode of every field. `StateEncoder`
serializes the whole state twice per block import, so this was pure wasted
work on the hot path.

## Performance comparison

Measured on the fuzz import path with picofuzz (3 reps per suite, median of
the 3 reported values). `baseline` = pre-change `main`; `change1` = commit 1
(types codec only); `change2` = + commit 3 (merklization + jamtests).

### CPU profile (picofuzz storage, steady state)

| Item | Before | After |
|------|--------|-------|
| `fmt.Sprintf` (cum) | 9.86% | gone (~0) |
| `types.(*Encoder).Encode` (cum) | 10.88% | gone |
| `fmt.(*pp).fmtBytes` (cum) | 7.48% | gone |
| total CPU samples | 2.94 s | 2.58 s (**-12%**) |

### Per-block mean latency (us)

| Suite | baseline | change1 | change2 |
|-------|----------|---------|---------|
| storage | 8934.2 | 7737.0 (-13.4%) | 7461.7 (**-16.5%**) |
| storage_light | 7268.7 | 6750.4 (-7.1%) | 6404.0 (-11.9%) |
| fallback | 5212.1 | 4771.4 (-8.5%) | 4590.3 (-11.9%) |
| safrole | 6862.0 | 6214.3 (-9.4%) | 6194.7 (-9.7%) |

### Per-block median latency (us)

| Suite | baseline | change1 | change2 |
|-------|----------|---------|---------|
| storage | 6689.1 | 5503.9 (-17.7%) | 5259.4 (**-21.4%**) |
| storage_light | 5375.1 | 4731.7 (-12.0%) | 4543.2 (-15.5%) |
| fallback | 2668.0 | 2142.3 (-19.7%) | 2084.5 (-21.9%) |
| safrole | 5673.4 | 5026.8 (-11.4%) | 5057.9 (-10.8%) |

change1 (the codec fix) provides most of the win; change2 adds a further
improvement on the storage suites (they go through the merklization state
parsing path). No extra runtime/RAM cost: escape analysis keeps the variadic
args off the heap (measured 0 heap boxings) and every suite got faster.

## Test plan

| Test | Result |
|------|--------|
| `go vet` (types, merklization, jamtests) | clean |
| `go test ./internal/types` (v0.7.2 vectors) | ok |
| `go test ./internal/utilities/merklization` | ok (after commit 5) |
| jam-test-vectors STF: preimages/reports/disputes/safrole/assurances | 8/42/28/21/10 pass, 0 fail |
| trace storage / storage_light | 100/100 pass |
| minifuzz x4 suites | 0 issues |
| picofuzz x4 suites | all "processed successfully" |
| jam-conformance traces (test_folder) | 1179 PASSED / 0 FAILED |

## Out of scope (follow-ups)

- Kept `var debugMode` (not `const`); escape analysis already removes the
  residual cost, so `const` showed no measurable gain.
- `jamtests/{accumulate,history,authorizations}` keep the old `cLog`
  signature (no eager `fmt.Sprintf` call sites, nothing to gain).

### DIFF
--- a/internal/types/decode.go
+++ b/internal/types/decode.go
@@ -14,7 +14,7 @@ func (h *HeaderHash) Decode(d *Decoder) error {
 	if err := binary.Read(d.buf, binary.LittleEndian, &val); err != nil {
 		return err
 	}
-	cLog(Yellow, fmt.Sprintf("HeaderHash: %x", val))
+	cLog(Yellow, "HeaderHash: %x", val)
 
 	*h = val
 	return nil
@@ -28,7 +28,7 @@ func (s *StateRoot) Decode(d *Decoder) error {
 	if err := binary.Read(d.buf, binary.LittleEndian, &val); err != nil {
 		return err
 	}
-	cLog(Yellow, fmt.Sprintf("StateRoot: %x", val))
+	cLog(Yellow, "StateRoot: %x", val)
 
 	*s = val
 	return nil
@@ -42,7 +42,7 @@ func (o *OpaqueHash) Decode(d *Decoder) error {
 	if err := binary.Read(d.buf, binary.LittleEndian, &val); err != nil {
 		return err
 	}
-	cLog(Yellow, fmt.Sprintf("OpaqueHash: %x", val))
+	cLog(Yellow, "OpaqueHash: %x", val)
 
 	*o = val
 	return nil
@@ -56,7 +56,7 @@ func (t *TimeSlot) Decode(d *Decoder) error {
 	if err := binary.Read(d.buf, binary.LittleEndian, &val); err != nil {
 		return err
 	}
-	cLog(Yellow, fmt.Sprintf("TimeSlot: %v", val))
+	cLog(Yellow, "TimeSlot: %v", val)
 
 	*t = val
 	return nil
@@ -113,13 +113,13 @@ func (e *EpochMark) Decode(d *Decoder) error {
 		return err
 	}
 
-	cLog(Yellow, fmt.Sprintf("Entropy: %x", e.Entropy))
+	cLog(Yellow, "Entropy: %x", e.Entropy)
 
 	if err = e.TicketsEntropy.Decode(d); err != nil {
 		return err
 	}
 
-	cLog(Yellow, fmt.Sprintf("TicketsEntropy: %x", e.TicketsEntropy))
+	cLog(Yellow, "TicketsEntropy: %x", e.TicketsEntropy)
 
 	// make the slice with validators count
 	e.Validators = make([]EpochMarkValidatorKeys, ValidatorsCount)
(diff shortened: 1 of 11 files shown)


yu2C (approved):
LGTM

yu2C (approved):

PR #1003, opened 2026-05-29
Fix/fuzz in memory persistence and validation ci

yu2C:
# PR: Fuzz target in-memory persistence + validation CI

**Base:** `v0.7.2.17` (`6a8d2773`)

## Summary

Long jam-conformance fuzz replays often ended with **client EOF**. Main causes:

1. **Unbounded `unfinalizedBlocks`:** `PruneOldData` ran only after a successful `ImportBlock`. On protocol-error paths the block was already `AddBlock`’d but the function returned early, so the in-memory chain could grow without limit → OOM.
2. **Disk growth:** `ResetInstance()` still opened **Pebble** via the persistent DB singleton; cross-trace `SetState` kept writing to disk and memory.
3. **Connection loss:** OOM, handler panic, or fatal errors surfaced as EOF on the client.

This PR under `JAM_FUZZ=1` uses **in-memory repositories only**, **trims unfinalized blocks to 24** after every `ImportBlock` (including protocol errors), and **recovers panics per connection** in the fuzz server so full conformance trees complete reliably.

Also adds a **Fuzz Validate** GitHub Actions workflow, local validation scripts, and English docs in `READMERef/`.

## Changed files

| File | Description |
|------|-------------|
| `internal/blockchain/chain_state.go` | Fuzz: shared memory `repo` / `persistentRepo`; `newChainState` / `ResetInstance`; `TrimUnfinalizedBlocksForFuzz`; skip disk persist/prune |
| `internal/fuzz/service.go` | `defer TrimUnfinalizedBlocksForFuzz()` on every `ImportBlock` |
| `internal/fuzz/server.go` | `serveOneRequest()` + per-request panic recover |
| `internal/fuzzenv/fuzzenv.go` | `FuzzPersistentRetainBlocks = 24` |
| `Makefile` | `validate-fuzz*` targets |
| `scripts/validate_fuzz.sh` | Local validation pipeline; target logs to `fuzz_target.log` |
| `scripts/ci_fuzz_docker_sock.sh` | Docker target + full conformance + fuzzy (CI) |
| `scripts/ci_jam_testing_full.sh` | FluffyLabs jam-testing (minifuzz + picofuzz) |
| `scripts/run_jam_testing_local.sh` | Optional local jam-testing smoke |
| `.github/workflows/fuzz-validate.yml` | Parallel CI jobs + summary gate |
| `READMERef/VALIDATE_FUZZ.md` | Validation spec (canonical doc) |
| `READMERef/PR_MANUAL_TEST_FUZZ_TARGET.md` | Manual socket test steps |
| `READMERef/INDEX.md` | Index entries for fuzz docs |
| `.gitignore` | Ignore bench artifacts, `.ci/`, `output*.txt` |

Intentional submodule bump: pkg/test_data/jam-test-vectors (<old> -> 95bd6f45) — required for accumulate gas tolerance and the traces/fuzzy set used by validate-fuzz.

Do not include in this PR: pkg/Rust-VRF submodule bump, scripts/bench-compare-fuzz-tags.sh, local notes (Todo.md, ci-pr.md).

## Behavior (before vs after)

| Item | `v0.7.2.17` | This PR |
|------|-------------|---------|
| Fuzz persistence | memory + **Pebble** | **memory only** (same in-memory DB for both repos) |
| `ResetInstance` | new memory + **new Pebble** | fresh in-memory `ChainState` |
| `unfinalizedBlocks` after protocol error | can **grow unbounded** | `defer` **KeepRecent(24)** |
| State / block writes | `persistentRepo` (disk) | **no disk** under fuzz |
| Server panic | can kill process | **recover**; close one connection |

## Performance / I/O (full tree, 1179 JSON)

Single fuzz server, one `test_folder` over  
`pkg/test_data/jam-conformance/fuzz-reports/0.7.2/traces`.  
Env: `JAM_FUZZ=1`, `JAM_FUZZ_SPEC=tiny`, `JAM_FUZZ_LOG_LEVEL=ERROR`.

| Version | Wall time | Result | Notes |
|---------|-----------|--------|-------|
| **v0.7.2.17** | **497 s** (~8m17s) | 1179/1179 PASS | Pebble path still used under fuzz |
| **This PR** | **390 s** (~6m30s) | 1179/1179 PASS | ~21.5% faster; no Pebble growth |

## CI: Fuzz Validate workflow

Jobs (parallel where possible):

1. **build-image** — `new-jamneration-target:ci` artifact  
2. **test-vectors** — jam-test-vectors (step 1)  
3. **test-trace** — trace modes (step 2)  
4. **fuzz-sock** — Docker target, full conformance + fuzzy traces  
5. **jam-testing** — minifuzz 6 + picofuzz 4 suites  
6. **fuzz-validate** — all jobs must succeed  

Spec: [READMERef/VALIDATE_FUZZ.md](READMERef/VALIDATE_FUZZ.md)

Path filters intentionally omit `pkg/**` and doc-only script edits so unrelated submodule/doc changes do not trigger the workflow.

## Verification (local)

- [x] Full conformance tree via socket — 1179/1179 PASS (`output.txt`)
- [x] Docker image + `test_folder` — 1179/1179 (~7m36s)
- [x] `make validate-fuzz-sock-smoke`
- [x] Benchmark vs `v0.7.2.17` (table above)

```bash
make build-target
make validate-fuzz-ci          # steps 1,2,3 + fuzzy
make validate-fuzz-sock-smoke    # quick conformance subset

TARGET_IMAGE=new-jamneration-target:ci ./scripts/ci_fuzz_docker_sock.sh
```

On sock failures, inspect `.jam_fuzz_docker_run/fuzz_target.log` (or `VALIDATE_FUZZ_TARGET_LOG`).

## Known limitations

- `SetState` / STF runtime / decode failures may still close the connection per protocol (not the main EOF fix).
- No dedicated unit test yet for the ≤24 unfinalized block cap.
- Legacy `./data/pebble` from pre-fix node runs can be deleted manually.

## Review focus

1. `newChainStateRepositories`: under fuzz, both repos always share the same memory DB.  
2. `ImportBlock`: `defer TrimUnfinalizedBlocksForFuzz` on all exit paths (including errors).  
3. `PersistStateForBlock` / `PruneOldData`: fuzz paths never write to `persistentRepo`.  
4. `serveOneRequest`: panic recover only affects the current connection.

## Test plan

- [ ] `Fuzz Validate` workflow green on GitHub (incl. `workflow_dispatch` once)  
- [ ] Optional: `./scripts/ci_jam_testing_full.sh` locally  
- [ ] Set `fuzz-validate` as required check after first green run

### DIFF
--- a/scripts/ci_jam_testing_full.sh
+++ b/scripts/ci_jam_testing_full.sh
@@ -1,0 +1,83 @@
+#!/usr/bin/env bash
+# Full minifuzz + picofuzz via FluffyLabs/jam-testing (npm/tsx), against a local target image.
+# See jam-testing .github/workflows/reusable-picofuzz.yml
+set -euo pipefail
+
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+
+TARGET_IMAGE="${TARGET_IMAGE:-new-jamneration-target:ci}"
+TARGET_NAME="${TARGET_NAME:-new-jamneration}"
+TARGET_CMD="${TARGET_CMD:-{TARGET_SOCK}}"
+TARGET_MEMORY="${TARGET_MEMORY:-2g}"
+JAM_TESTING_DIR="${JAM_TESTING_DIR:-${REPO_ROOT}/.ci/jam-testing}"
+JAM_TESTING_REF="${JAM_TESTING_REF:-main}"
+JAM_TESTING_REPO="${JAM_TESTING_REPO:-https://github.com/FluffyLabs/jam-testing.git}"
+
+# Cursor prompt 4 suites + official jam-testing extras (forks / no_forks)
+MINIFUZZ_SUITES="${MINIFUZZ_SUITES:-forks no_forks fallback safrole storage storage_light}"
+PICOFUZZ_SUITES="${PICOFUZZ_SUITES:-fallback safrole storage storage_light}"
+
+log() { printf '[ci_jam_testing] %s\n' "$*"; }
+die() { printf '[ci_jam_testing] ERROR: %s\n' "$*" >&2; exit 1; }
+
+command -v docker >/dev/null || die "docker required"
+command -v npm >/dev/null || die "npm required (Node 20+)"
+
+docker image inspect "${TARGET_IMAGE}" >/dev/null 2>&1 \
+	|| die "image not loaded: ${TARGET_IMAGE}"
+
+if [[ ! -d "${JAM_TESTING_DIR}/.git" ]]; then
+	mkdir -p "$(dirname "${JAM_TESTING_DIR}")"
+	log "cloning ${JAM_TESTING_REPO} (ref=${JAM_TESTING_REF}) -> ${JAM_TESTING_DIR}"
+	git clone --depth 1 --branch "${JAM_TESTING_REF}" "${JAM_TESTING_REPO}" "${JAM_TESTING_DIR}"
+fi
+
+(
+	cd "${JAM_TESTING_DIR}"
+	git fetch origin "${JAM_TESTING_REF}" --depth 1 2>/dev/null || true
+	git checkout "${JAM_TESTING_REF}" 2>/dev/null || git checkout "origin/${JAM_TESTING_REF}"
+	log "initializing jam-testing submodules (HTTPS; skip dashboard)"
+	# picofuzz-data nests jam-test-vectors via git@github.com — breaks non-interactive CI/WSL.
+	# forks/no_forks need picofuzz-conformance-data; picofuzz STF needs picofuzz-stf-data only (no nested jam-test-vectors).
+	git -c url.https://github.com/.insteadOf=git@github.com: \
+		submodule update --init --recursive picofuzz-conformance-data
+	git submodule update --init picofuzz-stf-data
+)
+
+log "npm ci + build client images in jam-testing"
+(
+	cd "${JAM_TESTING_DIR}"
+	npm ci
+	npm run build-docker -w @fluffylabs/picofuzz
+	npm run build -w @fluffylabs/minifuzz
+)
+
+export TARGET_NAME TARGET_IMAGE TARGET_CMD TARGET_MEMORY
+export TARGET_ENV="${TARGET_ENV:-}"
+
+run_tsx_test() {
+	local kind="$1"
+	local suite="$2"
+	local timeout_min="${3:-10}"
+	log "=== ${kind} ${suite} (timeout ${timeout_min}m) ==="
+	(
+		cd "${JAM_TESTING_DIR}"
+		TIMEOUT_MINUTES="${timeout_min}" npm exec tsx -- --test "tests/${kind}/${suite}.test.ts"
+	)
+}
+
+for suite in ${MINIFUZZ_SUITES}; do
+	timeout=10
+	[[ "$suite" == "storage" || "$suite" == "storage_light" ]] && timeout=15
+	run_tsx_test minifuzz "$suite" "$timeout"
+done
+
+for suite in ${PICOFUZZ_SUITES}; do
+	timeout=10
+	[[ "$suite" == "storage" || "$suite" == "storage_light" ]] && timeout=15
+	mkdir -p "${JAM_TESTING_DIR}/picofuzz-result"
+	chmod 777 "${JAM_TESTING_DIR}/picofuzz-result" 2>/dev/null || true
+	run_tsx_test picofuzz "$suite" "$timeout"
+done
+
+log "all jam-testing suites passed"
--- a/scripts/ci_jam_testing_full.sh
+++ b/scripts/ci_jam_testing_full.sh
@@ -27,15 +27,18 @@ docker image inspect "${TARGET_IMAGE}" >/dev/null 2>&1 \
 	|| die "image not loaded: ${TARGET_IMAGE}"
 
 if [[ ! -d "${JAM_TESTING_DIR}/.git" ]]; then
-	mkdir -p "$(dirname "${JAM_TESTING_DIR}")"
-	log "cloning ${JAM_TESTING_REPO} (ref=${JAM_TESTING_REF}) -> ${JAM_TESTING_DIR}"
-	git clone --depth 1 --branch "${JAM_TESTING_REF}" "${JAM_TESTING_REPO}" "${JAM_TESTING_DIR}"
+	mkdir -p "${JAM_TESTING_DIR}"
+	log "init ${JAM_TESTING_REPO} -> ${JAM_TESTING_DIR}"
+	git init -q "${JAM_TESTING_DIR}"
+	git -C "${JAM_TESTING_DIR}" remote add origin "${JAM_TESTING_REPO}"
 fi
 
 (
 	cd "${JAM_TESTING_DIR}"
-	git fetch origin "${JAM_TESTING_REF}" --depth 1 2>/dev/null || true
-	git checkout "${JAM_TESTING_REF}" 2>/dev/null || git checkout "origin/${JAM_TESTING_REF}"
+	# --depth 1 fetch of a specific commit works for SHA / tag / branch on GitHub.
+	log "fetch jam-testing @ ${JAM_TESTING_REF}"
+	git fetch --depth 1 origin "${JAM_TESTING_REF}"
+	git checkout -q FETCH_HEAD
 	log "initializing jam-testing submodules (HTTPS; skip dashboard)"
 	# picofuzz-data nests jam-test-vectors via git@github.com — breaks non-interactive CI/WSL.
 	# forks/no_forks need picofuzz-conformance-data; picofuzz STF needs picofuzz-stf-data only (no nested jam-test-vectors).
--- a/scripts/ci_jam_testing_full.sh
+++ b/scripts/ci_jam_testing_full.sh
@@ -10,7 +10,8 @@ TARGET_NAME="${TARGET_NAME:-new-jamneration}"
 TARGET_CMD="${TARGET_CMD:-{TARGET_SOCK}}"
 TARGET_MEMORY="${TARGET_MEMORY:-2g}"
 JAM_TESTING_DIR="${JAM_TESTING_DIR:-${REPO_ROOT}/.ci/jam-testing}"
-JAM_TESTING_REF="${JAM_TESTING_REF:-main}"
+# Pinned commit (not a moving branch) for reproducible CI; bump deliberately.
+JAM_TESTING_REF="${JAM_TESTING_REF:-028072cc1c40c3d2c39210aff67036e2018b2022}"
 JAM_TESTING_REPO="${JAM_TESTING_REPO:-https://github.com/FluffyLabs/jam-testing.git}"
 
 # Cursor prompt 4 suites + official jam-testing extras (forks / no_forks)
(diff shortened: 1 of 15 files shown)


YCC3741 (changes requested):
## Review: 

Solid, well-scoped fix with good docs and CI scaffolding. The `newChainState`/`newChainStateRepositories` extraction is a clean DRY win, the per-connection `recover()` in `serveOneRequest` is correct, and `INDEX.md` was updated per the docs policy. A few items should be resolved before merge — mainly a description/diff mismatch and an incomplete memory-bound guarantee on the protocol-error path.

---

### 🔴 Must resolve before merge

**1. PR description contradicts the diff (`pkg/test_data/jam-test-vectors` submodule bump)**
The description's "Do not include in this PR" lists *accidental submodule bumps (`jam-test-vectors`)*, but the diff bumps `pkg/test_data/jam-test-vectors` (`62959f4e → 95bd6f45`) and commit 3 bumps it explicitly. The new fuzzy validation reads `pkg/test_data/jam-test-vectors/traces/fuzzy`, so the bump is probably required.
→ Please confirm intent: if required, remove `jam-test-vectors` from the "do not include" list and keep the bump (verify the pinned commit is the intended v0.7.2 vectors); if accidental, drop it. An unintended bump can silently change the test baseline.

**2. Memory is not actually bounded on the protocol-error path** — `internal/fuzz/service.go` L51 / L143, `internal/blockchain/chain_state.go` L530
On a protocol error, `ImportBlock` returns before the success-path `PruneOldData` (L143). The new `defer cs.TrimUnfinalizedBlocksForFuzz()` (L51) only trims the `unfinalizedBlocks` slice via `KeepRecent(24)`; it does not remove the block that `AddBlock → persistBlockMapping` already wrote into the in-memory `repo`. `PruneOldData` reclaims `repo` entries only for blocks tracked in `persistedEntries`, which is appended on the success path only — so error-path blocks (header + extrinsic) are never reclaimed within a `ChainState` lifetime. `ResetInstance()` on each `SetState` bounds this per-trace, so it does not OOM on the current 1179-trace tree, but it does not honor the PR's headline "≤24 / bounded memory" invariant for long, error-heavy traces.
→ Either prune the just-added block from `repo` on the fuzz error path (reconcile `repo` with the retained window), or explicitly document this as a known limitation in `VALIDATE_FUZZ.md`.

---

### 🟡 Suggestions

**3. Validation can pass with zero tests run** — `scripts/validate_fuzz.sh` L122/L143 (and `ci_fuzz_docker_sock.sh` L21)
`run_test_folder_to_file` uses `... || true` and relies solely on `assert_no_failed_in_output` grepping `FAILED!!?:|FAILED:`. A mid-run target death *is* caught (the client prints `FAILED!!:` for the current and all remaining files, since it reuses one connection). But if the target dies before the handshake, or the folder matches zero JSON files, the client returns `error doing handshake: …` / walks nothing — output contains no `FAILED:`, the non-zero exit is swallowed by `|| true`, and the step passes with 0 traces tested. Note `testFolder` also `return nil` even when `failureCount > 0`, so grep is the only signal.
→ Assert an expected PASS count (e.g. number of traces) or `successCount > 0`, so a truncated/empty run can't be a false pass.

**4. jam-testing is pinned to a moving ref** — `scripts/ci_jam_testing_full.sh` L13
`JAM_TESTING_REF=main` means upstream `FluffyLabs/jam-testing` changes can break this gate unpredictably — which contradicts the "pin submodule, don't track main" policy in `VALIDATE_FUZZ.md`. Once `fuzz-validate` becomes a required check this blocks all PRs.
→ Pin a specific tag/commit and document the bump procedure.

**5. `pull_request` path filter omits `scripts/validate_fuzz.sh`** — `.github/workflows/fuzz-validate.yml` L12–L13
`pull_request.paths` lists `scripts/ci_*.sh` and `scripts/run_fuzz_target_docker.sh`, but not `validate_fuzz.sh` (which drives steps 1/2). `push.paths` correctly uses `scripts/**`. A future PR editing only `validate_fuzz.sh` would not trigger this gate.
→ Align with `push` (e.g. `scripts/**`).

**6. `cleanup_target` may leak the real server process** — `scripts/validate_fuzz.sh` L154
`kill "$FUZZ_TARGET_PID"` kills the `go run` parent; the compiled child server (holding the socket) often survives. The script also does not `rm -f "$FUZZ_SOCK"` before starting, so a stale/orphaned socket can break the next local run.
→ `go build` then run the binary directly, or start via `setsid` and `kill -- -$PGID`; remove the socket before bind.

**7. 15s startup timeout may be too short for a cold `go run`** — `scripts/validate_fuzz.sh` L20
`go run ./cmd/fuzz/` compiles before listening; 15s can be exceeded on a cold cache (local only — CI runs the sock path via the Docker script with 30s).
→ Raise the default or pre-build.

**8. Doc references a script that doesn't exist** — `READMERef/VALIDATE_FUZZ.md` L160
The "Pre-push local smoke" section tells readers to run `./scripts/ci_fuzz_pre_push.sh`, but that script is not in this PR (or the repo).
→ Add the script, or replace with the already-listed equivalent (`make validate-fuzz-ci` + `make fuzz-docker-build` + `./scripts/ci_fuzz_docker_sock.sh`).

**9. Dump target logs on CI failure** — `scripts/ci_fuzz_docker_sock.sh` L21
`assert_no_failed` dies without printing `docker logs "${TARGET_CONTAINER}"` (the local script tails its target log). CI failures are hard to diagnose without it.

**10. Collapse near-duplicate jobs** — `.github/workflows/fuzz-validate.yml` L88 / L123
`test-vectors` and `test-trace` differ only by `VALIDATE_FUZZ_STEPS=1` vs `2`; everything else (checkout, token, submodules, setup-go, rust, VRF build) is duplicated.
→ Use a `strategy.matrix`.

---

### 🔵 Nits

- **`chain_state.go` L809** — In the fuzz branch of `persistBlockMapping`, `repo.SaveBlock → SaveHeader` already stores the header→time-slot under the same hash, so the explicit `repo.SaveHeaderTimeSlot(...)` is redundant (harmless).
- **`chain_state.go` L71** — `getPersistentDatabase()` fuzz fallback returns a fresh, non-memoized `memory.NewDatabase()` per call. Currently unreachable (`newChainStateRepositories` returns first), but a future caller under fuzz would get a new empty DB each time. Consider memoizing or panicking to make misuse loud.
- **`server.go` L129** — Write errors now return → close the connection (previously `continue`d). Likely more correct; just confirm it's intended.
- **Script consistency** — `npm ci` vs `npm install`, `npm exec tsx --` vs `npx tsx`, "Node 20+" message while the workflow uses Node 24, and `JAM_FUZZ_HOST_DIR` (`ci_fuzz_docker_sock.sh`) vs `JAM_FUZZ_HOST_DATA` (`run_fuzz_target_docker.sh`) for the same bind-mount concept.
- **Commit hygiene** — Commit 3 combines `fix(fuzz): …` and `chore(submodule): …` in one subject; per `SEMANTIC_COMMIT_MESSAGES.md` the submodule bump should be its own `chore` commit.

---

### ✅ Looks good
- `GetInstance`/`ResetInstance` → `newChainState`/`newChainStateRepositories` removes duplication cleanly.
- Per-connection `recover()` in `serveOneRequest` with `%w` error wrapping.
- `READMERef/INDEX.md` updated for both new docs, per the documentation policy.

scripts/ci_jam_testing_full.sh
```diff
@@ -0,0 +1,83 @@
...
+set -euo pipefail
+
+REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+
+TARGET_IMAGE="${TARGET_IMAGE:-new-jamneration-target:ci}"
+TARGET_NAME="${TARGET_NAME:-new-jamneration}"
+TARGET_CMD="${TARGET_CMD:-{TARGET_SOCK}}"
+TARGET_MEMORY="${TARGET_MEMORY:-2g}"
+JAM_TESTING_DIR="${JAM_TESTING_DIR:-${REPO_ROOT}/.ci/jam-testing}"
+JAM_TESTING_REF="${JAM_TESTING_REF:-main}"
```
YCC3741: 🟡 **#4 (1/2)** — pin to a fixed commit instead of `main`. Note: `--branch` cannot take a raw SHA, so this needs the companion change to the clone block below.

```suggestion
# Pinned commit (not a moving branch) for reproducible CI; bump deliberately.
JAM_TESTING_REF="${JAM_TESTING_REF:-028072cc1c40c3d2c39210aff67036e2018b2022}"
```

scripts/ci_jam_testing_full.sh
```diff
@@ -0,0 +1,83 @@
...
+if [[ ! -d "${JAM_TESTING_DIR}/.git" ]]; then
+	mkdir -p "$(dirname "${JAM_TESTING_DIR}")"
+	log "cloning ${JAM_TESTING_REPO} (ref=${JAM_TESTING_REF}) -> ${JAM_TESTING_DIR}"
+	git clone --depth 1 --branch "${JAM_TESTING_REF}" "${JAM_TESTING_REPO}" "${JAM_TESTING_DIR}"
+fi
+
+(
+	cd "${JAM_TESTING_DIR}"
+	git fetch origin "${JAM_TESTING_REF}" --depth 1 2>/dev/null || true
+	git checkout "${JAM_TESTING_REF}" 2>/dev/null || git checkout "origin/${JAM_TESTING_REF}"
```
YCC3741: 🟡 **#4 (2/2)** — companion to the pinned SHA. `git fetch --depth 1 origin <sha>` works for a SHA / tag / branch on GitHub, whereas `git clone --branch <sha>` fails. Also drops `|| true` so a bad pin fails loudly instead of silently checking out the wrong ref.

```suggestion
if [[ ! -d "${JAM_TESTING_DIR}/.git" ]]; then
	mkdir -p "${JAM_TESTING_DIR}"
	log "init ${JAM_TESTING_REPO} -> ${JAM_TESTING_DIR}"
	git init -q "${JAM_TESTING_DIR}"
	git -C "${JAM_TESTING_DIR}" remote add origin "${JAM_TESTING_REPO}"
fi

(
	cd "${JAM_TESTING_DIR}"
	# --depth 1 fetch of a specific commit works for SHA / tag / branch on GitHub.
	log "fetch jam-testing @ ${JAM_TESTING_REF}"
	git fetch --depth 1 origin "${JAM_TESTING_REF}"
	git checkout -q FETCH_HEAD
```

YCC3741 (commented):
Re-reviewed after the rebase — thanks for the thorough pass; almost everything is addressed and CI is green. Two items from the original review are still open (both trivial). My earlier **Changes requested** still stands for these two.

**#1 — PR description still contradicts the diff.** The body still lists `jam-test-vectors` under "Do not include in this PR", but the diff intentionally bumps it. Suggested wording:

> **Intentional submodule bump:** `pkg/test_data/jam-test-vectors` (`62959f4e -> 95bd6f45`) — required for accumulate gas tolerance and the `traces/fuzzy` set used by validate-fuzz.
>
> **Do not include in this PR:** `pkg/Rust-VRF` submodule bump, `scripts/bench-compare-fuzz-tags.sh`, local notes (`Todo.md`, `ci-pr.md`).

**#4 — jam-testing still tracks `main`.** `scripts/ci_jam_testing_full.sh` still defaults `JAM_TESTING_REF` to `main` (no SHA anywhere in the script) and the workflow no longer overrides it, so the gate is not actually pinned yet. See the two inline suggestions.

Everything else looks good — #3 trace-count guard, #5 `scripts/**` path filter, #6 sock cleanup (setsid + `kill -- -PGID` + socket rm), #7 prebuilt binary, #8 `ci_fuzz_pre_push.sh`, #9 docker logs on failure, and #2 documented as a known limitation. 👍

YCC3741 (approved):

PR #1006, opened 2026-05-31
feat(up0): add leaf collection and announcement skip helpers

Terryhung:
Build a block index for ancestry queries, collect finalized descendants with no known children as leaves, and apply JAMNP-S UP 0 skip rules.


yu2C (approved):
LGTM.

PR #988, opened 2026-05-24
feat: globalKV refactoring + TrieStore + incremental merklize (Phase 1-2)

YCC3741:
## Summary

This PR implements the complete globalKV refactoring (Phase 1) and Trie Store with incremental merklize (Phase 2), delivering significant performance improvements for storage-heavy workloads.

### What was done

**Phase 1: globalKV Refactoring**
- Unified `StorageDict` + `LookupDict` into a single `globalKV map[StateKey][]byte` with O(1) counters
- Switched all PVM host calls (read/write/info/accumulate) to globalKV path
- Retired `unmatchedKeyVals` fallback pool, `ServiceAccountDerivatives`, dual-write legacy maps
- Fixed `ThresholdBalance` (GP 9.8) and `DeepCopy` correctness issues

**Phase 2.1: TrieStore Foundation**
- `internal/store/trie_store.go`: MerklizeAndCommit / MerklizeOnly / GetNode / GetNodeValue / TrieExists / refcount management / DeleteTrie
- Callback refactoring: `merklize()` / `merklizeWithCache()` now accept `storeNode` / `storeValue` callbacks
- 24 unit tests covering trie operations, refcount, shared nodes, callback consistency

**Phase 2.2: StateCommit Integration**
- Integrated TrieStore into `PersistStateForBlock` as the merklize path
- Eliminated triple computation in `ImportBlock` (3x serialize+merklize down to 1x)
- Added `lastCommittedStateRoot` field for O(1) prior state root access
- Trie lifecycle: ring buffer evict triggers DeleteTrie, ResetInstance triggers DeleteAll
- `restoreWithState` properly resets `recentStateRoots` + `lastCommittedStateRoot`

**Phase 2.3: Incremental Merklize**
- `internal/store/diff.go`: O(n) merge-scan producing inserted/deleted/modified entries
- `internal/store/incremental_merklize.go`: traverses prior trie along dirty paths, reuses unchanged subtrees
- Primary path with fallback: incremental when prior trie available, full merklize otherwise
- `KeyLevelCache.Invalidate(key)` on delete prevents stale cache hits
- Bug fixes: branch creation for degenerate cases, MSB mask correction in collapse

### Performance

| Test | Workload | Improvement |
|------|----------|-------------|
| **Minifuzz storage** | Every block has storage mutations | **-82.9%** |
| Minifuzz storage_light | Light storage | +2.5% |
| Minifuzz fallback/safrole | Consensus-heavy | +/-3% |
[...]

PR #1007, opened 2026-06-10
feat(telemetry): IsAuthorized/Refine cost sidecar to guarantor layer (#974 Phase 1a)

HanaYukii:
Phase 1 (guarantor half) of #974. **Stacked on #978** — retarget to `main` after it merges. Plumbing only: cost values stay zero until Phase 2a measures real numbers.

Carries the cost summaries for events **95 (Authorized)** and **101 (Refined)** from the PVM leaf to the guarantor layer, as a sidecar — never on consensus types:

```
PVM.Psi_I / RefineInvoke           Cost on transient return types
              │
WorkReportCompute ──▶ types.WorkReport            (consensus, untouched)
                  └─▶ WorkPackageTelemetryCost    (sidecar)
              │
Controller.TelemetryCost           ← #958 bridge will read here
```

**Design points (spec-driven)**
- `Refine` is per-item, index-aligned with `Items` — event 101's payload is one RefineCost per work item. Failed items keep their slot.
- Failure paths discard cost by design: JIP-3 maps guarantor failures to event 92 (reason string, no cost field); 95/101 fire on success only.
- `Process()` signature unchanged (sidecar lives on the controller) — auditing / CE134 callers untouched.

**Tests:** sentinel test plants distinct marker costs at the PVM leaf via a stub executor and asserts they reach the sidecar untouched, including a BadExports item keeping its slot. `work_package` / `pvmcost` / `telemetry` / `auditing` suites green (WSL); `PVM` package failures are pre-existing — byte-identical on the base commit.

**Not included:** event 47 (Accumulate) → Phase 1b (that chain crosses `ParallelizedAccumulation`, under active refactor in #988 follow-ups). Actual emission → #958.

### DIFF
--- a/internal/work_package/work_package_controller.go
+++ b/internal/work_package/work_package_controller.go
@@ -37,11 +37,13 @@ type WorkPackageController struct {
 	Bundle      []byte             // Shared
 
 	// TelemetryCost is the JIP-3 cost sidecar (#974) from the last
-	// successful Process call. Observability-only; zero before that, and
-	// left untouched when Process fails (JIP-3 failure paths map to event
-	// 92, which carries no cost — events 95/101 fire on success only).
-	// Not synchronized: read it only from the goroutine that called
-	// Process. The future event 95/101 bridge (#958) reads it here.
+	// fully-successful Process call — it is assigned only after every
+	// fallible step completes, so a failed Process never leaves a stale
+	// cost here. Observability-only; zero before the first success.
+	// (JIP-3 failure paths map to event 92, which carries no cost —
+	// events 95/101 fire on success only.) Not synchronized: read it only
+	// from the goroutine that called Process. The future event 95/101
+	// bridge (#958) reads it here.
 	TelemetryCost WorkPackageTelemetryCost
 }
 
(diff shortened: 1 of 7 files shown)


internal/work_package/work_package_controller.go
```diff
@@ -75,10 +83,11 @@ func (p *WorkPackageController) Process() (types.WorkReport, error) {
...
 
-	report, err := WorkReportCompute(&workPackage, p.CoreIndex, pa, pc, extrinsicMap, importSegments, delta, workPackageBundle, workPackageHash, p.PVM)
+	report, cost, err := WorkReportCompute(&workPackage, p.CoreIndex, pa, pc, extrinsicMap, importSegments, delta, workPackageBundle, workPackageHash, p.PVM)
 	if err != nil {
 		return types.WorkReport{}, err
 	}
+	p.TelemetryCost = cost
 	cs := blockchain.GetInstance()
 	newDict, err := cs.SetHashSegmentMapWithLimit(workPackageHash, types.OpaqueHash(report.PackageSpec.ExportsRoot))
 	if err != nil {
```
yu2C, line 94: `TelemetryCost` assigned before post-compute steps that can still fail
Maybe move `p.TelemetryCost = cost` to just before the final return report, nil (after segment lookup is attached)?

PR #1010, opened 2026-06-18
feat: jamnp up0 session

Terryhung:
## Summary
Implements JAMNP-S UP 0 session runtime: persistent per-peer stream handling, parallel handshake exchange, announcement read loop, and outbound AnnounceBlock with skip rules.

Builds on existing UP 0 wire codec (`wire.go`) and chain helpers (`chain.go`).

### DIFF
--- a/internal/networking/handler/up/chain.go
+++ b/internal/networking/handler/up/chain.go
@@ -33,6 +33,19 @@ func NewChainView(blocks []types.Block) (ChainView, error) {
 	return cv, nil
 }
 
+// ViewAtFinalized builds a chain view and returns the finalized block ref.
+func ViewAtFinalized(blocks []types.Block, finalized types.HeaderHash) (ChainView, BlockRef, error) {
+	cv, err := NewChainView(blocks)
+	if err != nil {
+		return ChainView{}, BlockRef{}, err
+	}
+	finalRef, ok := cv.hashToRef[finalized]
+	if !ok {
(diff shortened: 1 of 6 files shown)


yu2C (commented):
Thanks for the UP 0 session work. The core handler pieces are here and the package/networking tests pass, but I don’t think this is complete for #965 yet.

The main missing piece is duplicate UP stream handling. JAMNP-S requires that if multiple UP streams of the same kind are open, the stream with the greatest QUIC stream ID is kept and the others are reset/stopped:
https://github.com/zdave-parity/jam-np/blob/main/simple.md#L108-L114

Current `UP0Handler.registerSession` just overwrites the session by `peerKey` (`up0.go` L92–99), so it does not compare stream IDs or close the superseded stream. This leaves #965’s duplicate UP 0 stream acceptance criterion unmet.

A few other #965 gaps I’d expect to be addressed or explicitly scoped out:

- UP 0 is not wired into the peer stream dispatcher (`Peer.RegisterHandler(up.StreamKindUP0, ...)`).
- Grid-gossip eligibility is not enforced for opening UP 0 streams. JAMNP-S describes the UP 0 peer eligibility rules here: https://github.com/zdave-parity/jam-np/blob/main/simple.md#L192-L197
- Missing tests for duplicate stream replacement.
- Missing end-to-end test for `Handle()` / `readLoop`: Node A announces, Node B receives, `OnAnnouncement` fires.
- Missing negative validation test for invalid or non-descendant announcements.

PR #1011, opened 2026-06-23
docs: add Rust-VRF FFI usage documentation and update index

yu2C:
https://github.com/New-JAMneration/Rust-VRF/pull/28
- Introduced a new document detailing the usage patterns for the Rust-VRF FFI, covering handler lifecycle, ownership rules, FFI input slices, output buffer capacity, error handling, and secret material management.
- Updated the central index in READMERef/INDEX.md to include a link to the new Rust-VRF FFI usage document.


internal/safrole/sealing.go
```diff
@@ -159,6 +161,10 @@ func CalculateHeaderEntropy(public_key types.BandersnatchPublic, seal types.Band
 		(6.17) Hv ∈ F [] Ha ⟨XE ⌢ Y(Hs)⟩
 	*/
 	handler, _ := CreateVRFHandler(public_key)
```
YCC3741: **CRITICAL — `CalculateHeaderEntropy` returning `nil` panics at the call site**

This `nil` guard looks defensive, but the function result is converted to a fixed-size array by the caller:

```go
cs.GetProcessingBlockPointer().SetEntropySource(
	types.BandersnatchVrfSignature(CalculateHeaderEntropy(publicKey, seal)),
)
```

`BandersnatchVrfSignature` is `[BandersnatchSigSize]byte`. Since Go 1.20, converting a slice shorter than the array length to an array value triggers a runtime panic (`cannot convert slice with length 0 to array or pointer to array of length N`). So `return nil` does not make the flow safe — it merely swaps a nil-pointer-dereference panic for a slice-to-array conversion panic.

**Suggested fix:** return a zero-value signature, or change the signature to propagate an error so the caller handles the failure explicitly, instead of returning `nil`. For example:

```go
handler, err := CreateVRFHandler(public_key)
if err != nil || handler == nil {
	return make([]byte, types.BandersnatchSigSize)
}
defer handler.Free()
```

**Note:** strictly speaking this conversion panic already exists prior to this PR (when the error is ignored via `_`); the new nil-check does not actually fix it. Worth deciding here whether to address the root cause properly.

YCC3741 (approved):

PR #1023, opened 2026-06-27
feat(up0): wire UP 0 peer integration and duplicate stream handling

Terryhung:
## Summary

Wires UP 0 into the QUIC peer layer (follow-up to #965).

- **Duplicate UP streams:** `upStreamRegistry` keeps the greatest QUIC stream ID per `(peerKey, kind)`; lower IDs are reset before the handler runs (`peer.runUPStream`).
- **Peer wiring:** Node bootstrap registers `StreamKindUP0`; outbound UP 0 opens after `Connect` / inbound connection when `ShouldOpenUP0` allows.
- **Grid eligibility:** `ShouldOpenUP0` + `IsKnownValidator` added with unit tests. Runtime VM wiring deferred (`vm=nil` in bootstrap for now).
- **Fix:** `unregisterSession` only removes the session if it is still the active one.

Duplicate-stream logic lives in `quic/peer`, not `UP0Handler.registerSession`.

**Out of scope:** ValidatorManager from chain state, EventBus → CE 128, epoch connectivity gate.

### DIFF
--- a/internal/networking/quic/up_streams.go
+++ b/internal/networking/quic/up_streams.go
@@ -1,0 +1,81 @@
+package quic
+
+import (
+	"sync"
+
+	quicgo "github.com/quic-go/quic-go"
+)
+
+// upStreamKind is JAMNP-S UP 0 (block announcement).
+const upStreamKind byte = 0
+
+type upStreamRegistry struct {
+	mu     sync.Mutex
+	active map[string]map[byte]trackedUPStream
+}
+
+type trackedUPStream struct {
+	id     quicgo.StreamID
+	stream quicgo.Stream
+}
+
+func newUPStreamRegistry() *upStreamRegistry {
+	return &upStreamRegistry{
+		active: make(map[string]map[byte]trackedUPStream),
+	}
+}
+
+// admit records a UP stream and returns whether the caller should run the handler.
+// Per JAMNP-S, only the UP stream with the greatest QUIC stream ID is kept.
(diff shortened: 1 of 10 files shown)


internal/networking/quic/up_streams.go
```diff
@@ -0,0 +1,81 @@
...
+	if len(byKind) == 0 {
+		delete(r.active, peerKey)
+	}
+}
+
+func cancelStream(stream quicgo.Stream) {
+	if stream == nil {
+		return
+	}
+	_ = stream.CancelRead(0)
```
yu2C: This would cause a compile error, so remove `_ =`.

yu2C (commented):
Two things to fix before merge:

1. OnAnnouncement not wired → EventBus

registerUP0Handler registers the UP0 handler but never sets OnAnnouncement. Valid announcements stop inside the handler — nothing gets published to EventBus, so SyncManager never sees new block headers.

Need something like:
```
up0.OnAnnouncement = func(ann uphandler.Announcement, pk ed25519.PublicKey) error {
    // publish PeerUpdated (or similar) via eventBus
    return nil
}
```
This is #965 scope (hook only, no block import). CE 128 is #966.

2. up_streams_test.go — wrong assertion

In TestUPStreamRegistryKeepsGreatestStreamID, when high (id=8) replaces low (id=4), admit calls cancelStream(low), so low.canceled should be true. Line 45 expects false — flip it:
```
require.True(t, low.canceled)
```

yu2C (approved):

PR #1025, opened 2026-06-28
feat(safrole): dynamic ticket cap + per-block K cap (GP v0.8.0) (#1013)

HanaYukii:
Part of #1012 · Closes #1013

## The behaviour change: dynamic ticket cap (`eq:ticketsextrinsic`)

A validator's ticket entry-index cap goes from a **fixed** value to **computed**:

```
before:  Attempt < 3            (fixed TicketsPerValidator)
after:   Attempt < n,  n = ⌈2E / |γ'_K|⌉
```

`E` = epoch length, `γ'_K` = next-epoch validator sequence. **tiny: n = ⌈24/6⌉ = 4** (entry-index 3 is now valid; was rejected). Full: n = 2. Fewer validators ⇒ larger n, so the ticket pool can still fill.

This is the **only** consensus behaviour change in the PR.

## Also (alignment, not new behaviour)

`VerifyEpochTail`: per-block ticket count capped by `MaxTicketsPerBlock` (K), not `ValidatorsCount` (V) — `eq:enforceticketlimit`. The comment already said K; the code used V.

`VerifyTicketsAttempt`: the defensive `|γ'_K| == 0` branch now returns `nil` for an empty extrinsic (nothing to reject) rather than `bad_ticket_attempt`.

`eq:valcount` and the slot-sealer family: **no change** (V values 6/1023 already satisfy `{3c}`; transitions already match).

## Tested

- ✅ `TestTicketsAttemptDynamicCap` (new) — proves `n` tracks `|γ'_K|`.
- ✅ `TestVerifyEpochTail` — K+1 pre-tail regression.
- ✅ Negative case bumped `Attempt 3 → 4` (3 is valid under n=4).
- ✅ `go test ./internal/safrole/...` + CI green.
- ⏭️ `publish-tickets-no-mark-1` (v0.7.2 vector that expects entry-index 3 to fail) is skipped through **one shared list**: `testdata.IsV080IncompatibleVector`, consumed by **both** read paths — `ReadTestData` (the `cmd/node test` path) and `TestSafroleTestVectors`. One entry, removed from both paths together once official v0.8.0 vectors land.

## Out of scope
Host-call CONSTS still encode `N` (v0.8.0 drops it) — separate serialization issue. `VERSION_GP` bump — separate.

### DIFF
--- a/internal/safrole/extrinsic_tickets.go
+++ b/internal/safrole/extrinsic_tickets.go
@@ -128,10 +128,15 @@ func VerifyTicketsDuplicate(tickets types.TicketsAccumulator) *types.ErrorCode {
 func VerifyTicketsAttempt(tickets types.TicketsExtrinsic) *types.ErrorCode {
 	numV := len(blockchain.GetInstance().GetPosteriorStates().GetGammaK())
 	if numV == 0 {
-		// Defensive only: γ'_K is never empty in this codebase (offenders
-		// are zeroed in place). Avoid a divide-by-zero if that ever changes.
-		err := SafroleErrorCode.BadTicketAttempt
-		return &err
+		// Defensive only: γ'_K is never empty in this codebase (offenders are
+		// zeroed in place). With no validators the cap n is undefined, so reject
+		// only if there is actually a ticket to bound; an empty extrinsic has
+		// nothing to reject. Also avoids a divide-by-zero if this ever changes.
+		if len(tickets) > 0 {
+			err := SafroleErrorCode.BadTicketAttempt
+			return &err
+		}
+		return nil
 	}
 	// n = ceil(2E / numV) via integer arithmetic.
 	n := (2*types.EpochLength + numV - 1) / numV
--- a/internal/safrole/extrinsic_tickets.go
+++ b/internal/safrole/extrinsic_tickets.go
@@ -1,4 +1,4 @@
-// 6.7. The Extrinsic and Tickets (graypaper 0.5.4)
+// 6.7. The Extrinsic and Tickets (graypaper 0.8.0)
 package safrole
 
 import (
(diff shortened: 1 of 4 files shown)


YCC3741 (changes requested):
Nice work on this one — I verified the dynamic cap against GP v0.8.0 (`n = ceil(2E / |γ'_K|)`, `eq:ticketsextrinsic`, gavofyork/graypaper #514). The core logic, the STF ordering (`KeyRotate` sets `γ'_K` before `VerifyTicketsAttempt`), and the `K` cap in `VerifyEpochTail` all look correct. Just two small things before merge:

**DRY — `safrole_test.go` & `testdata/reader.go`: the skip knowledge is duplicated**

The "`publish-tickets-no-mark-1` is v0.8.0-incompatible" skip lives in two spots via two mechanisms: the inline `if binFile == "publish-tickets-no-mark-1.bin"` in `TestSafroleTestVectors`, and the `v080IncompatibleVectors` map in `testdata/reader.go` (consumed by `cmd/node/test.go` through `ReadTestData`). Both read paths are genuinely separate, so both skips are needed — but the PR description's "shared reader skip list" doesn't quite match the test path, since `TestSafroleTestVectors` doesn't consume the shared list.

Could we either have both paths consume one exported shared list, or at least cross-reference the two spots in their comments? That way the eventual cleanup (once official v0.8.0 vectors land) removes both entries together rather than risking a leftover.

**NIT — `extrinsic_tickets.go`: `numV == 0` guard on an empty extrinsic**

When `numV == 0` and `tickets` is empty, returning `nil` would be a touch more accurate than `BadTicketAttempt` (there's nothing to reject). It's unreachable in practice since `γ'_K` is never empty, so this is entirely optional — gating on `len(tickets) > 0` would cover it.

Thanks!

PR #1028, opened 2026-06-30
feat: node ce128 sync

Terryhung:
## Summary

Follow-up to merged UP 0 peer work (#965). This PR closes the remaining bootstrap/sync gaps that were previously scoped out.

**Done**
- **ValidatorManager from chain** — `NewValidatorManagerFromChain` builds grid state from `Lambda` / `Kappa` / `GammaK`; validator nodes pass it into `ShouldOpenUP0` so grid eligibility is enforced at runtime (full nodes still open UP 0 to all peers).
- **CE 128 client** — `ce.RequestBlocks` opens a CE 128 stream (`WriteStreamKind(128)` → framed request → read framed block responses).
- **SyncManager block fetch** — on `PeerUpdated` with a new header (from UP 0 `OnAnnouncement`), requests missing blocks via CE 128 and stores them with `ChainState.AddBlock`; `bulkSync` uses the same path.
- **Peer lookup** — `RemotePeer`, `ConnectionFor`, `PeerSet.GetByEd25519`; announcements resolve the tracked peer and set `Best` before publishing.

**Not in this PR**
- Full **validated block import** (#966) — no consensus/import pipeline; blocks are fetched and appended via `AddBlock` only.
- **Multi-round bulk sync** — single CE 128 request per sync step; does not loop until `networkBest` is reached.
- **Dual-node QUIC E2E** for sync; **malformed announcement** wire tests.
- Epoch-transition connectivity gate.

### DIFF
--- a/internal/networking/handler/ce/ce128_client.go
+++ b/internal/networking/handler/ce/ce128_client.go
@@ -1,0 +1,58 @@
+package ce
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+
+	"github.com/New-JAMneration/JAM-Protocol/internal/networking/quic"
+	"github.com/New-JAMneration/JAM-Protocol/internal/types"
+)
+
+// RequestBlocks opens a CE 128 stream on conn and returns the requested blocks.
+func RequestBlocks(ctx context.Context, conn *quic.Connection, req CE128Payload) ([]types.Block, error) {
+	if conn == nil {
+		return nil, fmt.Errorf("nil connection")
+	}
+
+	qstream, err := conn.OpenStreamSync(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("open CE 128 stream: %w", err)
+	}
+	stream := &quic.Stream{Stream: qstream}
+
+	if err := stream.WriteStreamKind(byte(BlockRequest)); err != nil {
+		_ = stream.Close()
+		return nil, fmt.Errorf("write stream kind: %w", err)
+	}
+
+	handler := NewDefaultCERequestHandler()
+	payload, err := handler.Encode(BlockRequest, &req)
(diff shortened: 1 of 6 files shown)


internal/networking/handler/ce/ce128_client.go
```diff
@@ -0,0 +1,58 @@
...
+func RequestBlocks(ctx context.Context, conn *quic.Connection, req CE128Payload) ([]types.Block, error) {
+	if conn == nil {
+		return nil, fmt.Errorf("nil connection")
+	}
+
+	qstream, err := conn.OpenStreamSync(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("open CE 128 stream: %w", err)
+	}
+	stream := &quic.Stream{Stream: qstream}
```
yu2C, line 23: `defer stream.Close()`

internal/networking/handler/ce/ce128_client.go
```diff
@@ -0,0 +1,58 @@
...
+		return nil, fmt.Errorf("encode CE 128 request: %w", err)
+	}
+	if err := stream.WriteMessage(payload); err != nil {
+		_ = stream.Close()
+		return nil, fmt.Errorf("write CE 128 request: %w", err)
+	}
+
+	decoder := types.NewDecoder()
+	blocks := make([]types.Block, 0, req.MaxBlocks)
+	for {
```
yu2C: `for len(blocks) < int(req.MaxBlocks)`
Suggest breaking once over upper bound. per https://github.com/New-JAMneration/JAM-Protocol/blob/5289f5e3b0d3a975c1f3b3bad5fa180212a34afd/internal/networking/handler/ce/ce128.go#L22

yu2C (approved):

PR #1031, opened 2026-07-02
feat(header): BlockInfo timeslot + epoch-mark length prefix + preimage-hash extrinsic commitment (GP v0.8.0) (#1014)

HanaYukii:
Part of #1012 · Closes #1014

> **Stacked on #1027** (← #1026). Base is `feat/reporting-1016-workcontext-fields`; review only the #1014 commit here. Retarget to `1012-update-to-v080` as the stack merges.

## What — three consensus-visible v0.8.0 changes

| # | change | GP ref |
|---|--------|--------|
| 1 | **BlockInfo** (recent history) gains a 4-byte `timeslot` between `state_root` and `reported`; the appended item carries `H_t` | `eq:recenthistoryspec`, `eq:recenthistorydef`, C(3) |
| 2 | **EpochMark**'s validator-key sequence is now **length-prefixed** (`var{k}`); was fixed-length | `encodeepochmark` (serialization.tex) |
| 3 | **Extrinsic hash**: preimages component becomes `p = E(var[(E4(s), blake(d))])` — each blob committed by its Blake2b hash, replacing the full C.15 encoding | header.tex |

Everything else in the issue is **already aligned** (verified against the v0.7.2→v0.8.0 tex diff): epoch/winning-tickets marker presence+content rules, the ancestors requirement, and the author-index bound are semantically unchanged.

## Changes

- `types.BlockInfo` + Encode / Decode / UnmarshalJSON (+ jam_types struct kept in sync; reflection registries stay on v0.7.x per #1015 precedent)
- `EpochMark.Encode/Decode` — emit/read the length prefix (decode still validates it equals `ValidatorsCount`)
- `block_serialization.go` — new `p()` helper (mirrors `g()`) used by `CreateExtrinsicHash`; **full-block serialization (C.15) unchanged** — only the hash commitment changes
- `recent_history.NewItem` takes the block slot; both STF call sites pass `block.Header.Slot`

## Tested

- ✅ `TestEncodeBlockInfo_V080Timeslot`, `TestEncodeEpochMark_V080LengthPrefix` (new) — round-trip + byte-offset asserts
- ✅ `TestExtrinsicHashPreimagesComponent_V080` (new) — count prefix + E4 requester + blake(blob), empty case
- ✅ `go build` / `go vet` / `gofmt` clean; `internal/types`, `recent_history`, `safrole`, `statistics`, `store`, `extrinsic`, `input/jam_types` green
[...]

PR #1027, opened 2026-06-28
feat(reporting): add anchor_slot + lookup_anchor_state_root to RefineContext (GP v0.8.0) (#1016)

HanaYukii:
Part of #1012 · Closes #1016

> **Stacked on #1026** (#1015). Base is `feat/workpackage-1015-erasure-shards`; review only the two #1016 commits here. Retarget to `1012-update-to-v080` once #1026 merges.

### 1. RefineContext: 6 → 8 fields (`eq:workcontext`)

| + field | type | position |
|---------|------|----------|
| `anchor_slot` | `u32` (4B) | right after `anchor` |
| `lookup_anchor_state_root` | hash (32B) | right after `lookup_anchor_slot` |

Final order: `anchor, anchor_slot, state_root, beefy_root, lookup_anchor, lookup_anchor_slot, lookup_anchor_state_root, prerequisites`. Consensus-visible (RefineContext rides inside every WorkReport / WorkPackage).

### 2. ρ stores the full guarantee (`eq:reportingstate` / C(10)) — added after a follow-up verification pass

v0.7.x stored `(workreport, timestamp)` per core; v0.8.0 stores **`(guarantee, timestamp)`** — the whole guarantee (report + slot + credential). Wire: `report ++ E4(slot) ++ var(credential) ++ E4(timestamp)`.

- `AvailabilityAssignment.Report WorkReport` → `Guarantee ReportGuarantee` + codec/JSON
- Writer `TransitionWorkReport` stores the full guarantee (it already had it in hand)
- Readers updated mechanically (`.Report` → `.Guarantee.Report`): assurance availability/timeout filtering, guarantee dedup, disputes `ClearWorkReports` hashing, auditing queue selection

## Notes

- **Reflection-codec registries + jam_types mirror left on v0.7.x** (same as #1015): the reflection schema is name-keyed, so struct and schema must migrate together when official v0.8.0 vectors land. For RefineContext the mirror struct gained the two extra fields (schema ignores extras); for ρ the shape change must wait.
- JSON key `"guarantee"` pends confirmation by the official v0.8.0 vector schema.

## Tested

- ✅ `TestEncodeRefineContext_V080Fields` (new) — round-trip + byte-offset asserts
- ✅ `TestEncodeAvailabilityAssignment_V080Guarantee` (new) — asserts the credential + guarantee slot now sit between report and timestamp; round-trips
[...]

PR #1026, opened 2026-06-28
feat(work-package): add erasure_shards to WorkPackageSpec (GP v0.8.0) (#1015)

HanaYukii:
Part of #1012 · Closes #1015 · base = `1012-update-to-v080`

## The change: WorkPackageSpec gains a serialized erasure_shards field

GP v0.8.0 `eq:avspec` inserts a 2-byte `erasure_shards` (u16) into the availability specifier, between `erasure_root` and `seg_root`:

```
v0.7.x:  hash, length, erasure_root,                exports_root, exports_count   (5 fields)
v0.8.0:  hash, length, erasure_root, erasure_shards, exports_root, exports_count   (6 fields)
```

This is **consensus-visible**: `package_spec` rides on every on-chain WorkReport, so the byte layout of every report/guarantee/assurance changes. Verified twice against the GP `serialization.tex` (`\encode[2]{erasureshards}`).

## What changed

- **`WorkPackageSpec`** (`types.go`): `+ ErasureShards U16`. Encode / Decode / UnmarshalJSON and the reflection codec (`input/jam_types`) all emit it in spec order.
- **`A()` producer** (`work_package.go`): sets `ErasureShards = TotalShards` (the shard count the bundle is erasure-coded into).
- **Comment-only**: segment-footprint log2 arg relabelled `W_X` (was `W_M`; value 4488 unchanged); `make-bundle` / `compute-report` / `availability-specifier` GP labels updated to v0.8.0 eq names.

## No change

Limit constants (W_F, W_B, I, T, G_A …) are all unchanged v0.7.2 → v0.8.0.

## Tested

- ✅ `TestEncodeWorkPackageSpec_ErasureShards` (new) — round-trip + asserts the 2-byte shard count sits immediately after the 32-byte erasure root.
- ✅ `go test ./internal/types/ ./internal/work_package/` green.
- ⏭️ Vendored jam-test-vectors are still v0.7.x (5-field `package_spec`) so they no longer decode — skipped until official v0.8.0 vectors land (#1012 non-goal):
  - `internal/types` decode/encode WorkReport vectors → `skipV080VectorIncompat`.
  - `cmd/node test` path: reports / assurances / disputes / accumulate modes skipped whole via `v080IncompatibleModes` (statistics kept — its 1/3-pass expectation is unaffected).

PR #1032, opened 2026-07-02
feat(disputes): verdict judgment length prefix + extrinsic caps + drop culprits rule (GP v0.8.0) (#1017)

HanaYukii:
> ## ⚠️ Behavioural change — not just a wire tweak
> Beyond the length-prefix, this **deletes** a v0.7.x validation rule ("bad verdict ⇒ ≥2 culprits") and **adds** extrinsic-size rejections (verdicts/offenses ≤ 16). Both change **which disputes extrinsics are accepted vs rejected** — please review the accept/reject logic, not only the byte layout.

Part of #1012 · Closes #1017

> **Stacked on #1031** (← #1027 ← #1026). Base is `feat/header-1014-recenthistory-v080`; review only the #1017 commit here. Retarget as the stack merges.

## What — three v0.8.0 changes

| # | change | GP ref |
|---|--------|--------|
| 1 | **Wire**: each verdict's judgment sequence is now **length-prefixed** (`var{...}`); was a fixed `⌊2V/3⌋+1` entries. Decode still validates the count (5 tiny / 683 full) | `encodedisputes` (serialization.tex) |
| 2 | **Extrinsic caps** (behavioural): `|verdicts| ≤ 16`, `|culprits|, |faults| ≤ 16` — new constants + rejection, previously unenforced | `eq:disputesextrinsics`, `Cmaxextrinsicverdicts` / `Cmaxextrinsicoffenses` |
| 3 | **Rule removed** (behavioural): v0.8.0 keeps only "good verdict ⇒ ≥1 fault"; the v0.7.x "bad verdict ⇒ ≥2 culprits" rule is **deleted** — `ValidateCulprits` and its call site removed (the issue text didn't mention this deletion) | judgments.tex v0.7.2→v0.8.0 diff |

**Already aligned** (verified against the tex diff, no change): good/bad/wonky set updates, offender tracking (`eq:offendersdef` is a LaTeX re-flow only), rho clearing, age ∈ {e, e−1}, active-vs-previous key-set selection, judge-index bounds, sort/uniqueness rules. The κ-vs-V threshold re-expression is a no-op here (offender keys zeroed in place ⇒ |κ| ≡ V).

## Notes

- `NotEnoughCulprits` keeps its ordinal (conformance error mapping) but is no longer produced; `too_many_verdicts` / `too_many_offenses` ordinals are provisional until the official v0.8.0 error enum ships.
- Found in passing: `TestDisputes` pointed at a non-existent vector dir and **silently ran zero vectors**. Path fixed to `stf/disputes` with a fail-on-empty guard, then skipped (v0.7.x vectors predate both the wire change and the culprits-rule removal).

## Tested

- ✅ `TestEncodeVerdict_V080LengthPrefix` (new) — round-trip + prefix at byte offset 36, total length assert
- ✅ `go build` / `go vet` / `gofmt` clean; `internal/extrinsic`, `internal/types` green
- ⏭️ v0.7.x disputes vectors skipped (wire + expected-output incompatible) until official v0.8.0 vectors land (#1012)


internal/extrinsic/dispute.go
```diff
@@ -10,6 +10,19 @@ func Disputes() (types.OffendersMark, error) {
 	block := blockchain.GetInstance().GetLatestBlock()
 	disputeExtrinsic := block.Extrinsic.Disputes
 
+	// GP v0.8.0 eq:disputesextrinsics caps the extrinsic sequences:
+	// |verdicts| <= Cmaxextrinsicverdicts, |culprits|,|faults| <=
+	// Cmaxextrinsicoffenses (both 16).
+	if len(disputeExtrinsic.Verdicts) > types.MaxExtrinsicVerdicts {
```
yu2C: **[SUGGESTION] Enforce extrinsic caps in `DisputesExtrinsic.Validate()`**

GP v0.8.0 caps (`|verdicts| ≤ 16`, `|culprits|`, `|faults| ≤ 16`) are checked inline here. Consider moving them into the existing `DisputesExtrinsic.Validate()` in `internal/types/types.go` instead:

```go
func (d *DisputesExtrinsic) Validate() error {
	if len(d.Verdicts) > MaxExtrinsicVerdicts {
		return errors.New("too_many_verdicts")
	}
	if len(d.Culprits) > MaxExtrinsicOffenses || len(d.Faults) > MaxExtrinsicOffenses {
		return errors.New("too_many_offenses")
	}
	for _, verdict := range d.Verdicts {
		if err := verdict.Validate(); err != nil {
			return err
		}
	}
	return nil
}
```

`ScaleDecode` already calls `Validate()` after decode, so oversized extrinsics would fail at the type layer. `Disputes()` could then call `disputeExtrinsic.Validate()` (mapping the error string) rather than duplicating length checks.

internal/extrinsic/dispute_test.go
```diff
@@ -81,8 +81,20 @@ func GetTestJsonFiles(dir string) []string {
 }
 
 func TestDisputes(t *testing.T) {
-	dir := filepath.Join(JAM_TEST_VECTORS_DIR, "disputes", types.TEST_MODE)
+	// The v0.7.x disputes vectors predate GP v0.8.0: the verdict judgment
+	// sequence is now length-prefixed (encodedisputes var{...}, #1017) and the
+	// ">= 2 culprits per bad verdict" rule (not_enough_culprits) was dropped,
+	// so both the .bin layout and several expected outputs no longer hold.
+	// Re-enable on official v0.8.0 vectors (#1012).
+	t.Skip("v0.7.x disputes vectors predate GP v0.8.0 (#1017); re-enable on official v0.8.0 vectors")
```
yu2C, line 90: **[SUGGESTION] Add unit tests for cap rejection**

No test currently covers the `too_many_verdicts` / `too_many_offenses` paths. A small table-driven test (e.g. 17 verdicts / 17 culprits / 17 faults) would lock behaviour once caps live in `Validate()` or `Disputes()`.

yu2C (approved):

PR #1033, opened 2026-07-02
fix(accumulation): eq:accseq gas budget + processed-transfers output (GP v0.8.0) (re #1019)

HanaYukii:
> ## ⚠️ CONSENSUS-CRITICAL — please review the math, not just the diff
> This is **not** a notation refactor. It changes the accumulation gas budget (`eq:accseq`): **how many work-reports accumulate per block**, plus the `g*` recursion term. A wrong value **diverges the posterior state root** on any block carrying deferred transfers or free-accumulation services. Check the gas math against `accumulation.tex` at tag **v0.8.0** directly.

Part of #1012 · Re #1019 (closed as no-change — this PR covers three `eq:accseq` deltas found in a later verification pass; see the comment on #1019)

> **Stacked on #1032** (← #1031 ← #1027 ← #1026). Base is `feat/disputes-1017-v080`; review only this commit. Retarget as the stack merges.

## What — three v0.8.0 changes to `eq:accseq` (∆+)

| # | change | v0.7.2 | v0.8.0 |
|---|--------|--------|--------|
| 1 | **Prefix selection budget** (consensus-critical: how many reports accumulate per block) | `Σ digest gas ≤ g` | `Σ digest gas + Σ incoming-transfer gas + Σ values(f) ≤ g` |
| 2 | **g\*** transfer-gas term | sums the **incoming** `t` | sums the transfers **produced this round** (`t*`) |
| 3 | **Return arity** | 4-tuple | 5-tuple: adds **processed transfers** `t ⌢ t†` |

Verified directly against `accumulation.tex` at both tags. Change 3 is required by #1021 (statistics `T(s)` = transfers processed per destination service).

## Changes

- `accumulationPrefixLen()` — selection rule extracted + fixed, with a table-driven unit test (transfer/free-gas reservation, exact-fit, exhausted-budget cases)
- `OuterAccumulation` — `g*` now sums `t*`; `ProcessedTransfers` threaded through base case and combine step
- `OuterAccumulationOutput.ProcessedTransfers` — new field (unconsumed until #1021 lands)

## Tested

- ✅ `TestAccumulationPrefixLen_V080GasBudget` (new, 5 cases)
- ✅ `go test ./internal/accumulation/` green; `go build` / `go vet` / `gofmt` clean
- ⏭️ `TestAccumulateTestVectors` now skipped: the v0.7.x vectors predate the #1015/#1016 wire changes and decoding them with the v0.8.0 decoder mis-parses a length prefix and **OOMs** (pre-existing on the base — the CI path already skips the mode). Flagged separately: the decoder should bound length prefixes instead of trusting them.


YCC3741 (commented):
## Summary

The implementation matches the GP v0.8.0 `eq:accseq` formula, including the expanded prefix budget, the use of produced transfers in `g*`, and the processed-transfer output. The main concerns are missing direct regression coverage for two consensus-critical behaviors, an unconditional conformance-test skip, and branch history that conflicts with the repository's rebase policy.

**Stats:**
- :red_circle: Critical: 0
- :yellow_circle: Suggestion: 2


## Files Reviewed

| File | Status | Comments |
|------|--------|----------|
| `internal/accumulation/accseq_v080_test.go` | :yellow_circle: Suggestions | 1 |
| `internal/accumulation/accumulation.go` | :white_check_mark: Clean | 0 |
| `internal/accumulation/accumulation_test.go` | :yellow_circle: Suggestions | 1 |
| `internal/accumulation/types.go` | :white_check_mark: Clean | 0 |

---

## File: `internal/accumulation/accseq_v080_test.go`

### :yellow_circle: SUGGESTION — L12: Cover the other two consensus-critical `eq:accseq` changes

**Problem:**
This test only exercises prefix selection. The PR also changes `g*` to use transfers produced by the current round and adds the recursive `t concat t-dagger` output, but neither behavior has direct regression coverage. This is especially important because the existing accumulation vectors are now skipped.

**Current code (L9-L12):**
```go
// TestAccumulationPrefixLen_V080GasBudget covers the GP v0.8.0 eq:accseq
// prefix-selection budget: sum(digest gas) + sum(incoming transfer gas) +
// sum(free-accumulation gas) <= g. Under v0.7.x only the digest gas counted.
func TestAccumulationPrefixLen_V080GasBudget(t *testing.T) {
```

**Suggested improvement:**
Add focused `OuterAccumulation` tests that:

1. Make one accumulation round produce a deferred transfer and verify that the next round's budget adds the produced transfer's gas, rather than the incoming transfer's gas.
2. Use at least two recursive rounds and assert that `ProcessedTransfers` preserves the exact `t concat t-dagger` order.

**Why:** A regression in either path changes accumulation selection or statistics and can therefore diverge the posterior state root.

---

## File: `internal/accumulation/accumulation_test.go`

### :yellow_circle: SUGGESTION — L328: Avoid permanently disabling accumulation vectors

**Problem:**
The unconditional skip remains active even after the test-vector submodule is updated to official v0.8.0 vectors. Nothing automatically alerts maintainers to remove it, so accumulation conformance coverage can remain disabled indefinitely.

**Current code (L323-L329):**
```go
// The v0.7.x accumulate vectors carry WorkReports whose wire layout
// predates GP v0.8.0 (#1015/#1016), so the decoder mis-parses a length
// prefix and dies on a giant allocation (OOM). Re-enable on official
// v0.8.0 vectors (#1012); the CI path already skips the mode via
// testdata.v080IncompatibleModes.
t.Skip("v0.7.x accumulate vectors predate GP v0.8.0 wire changes (#1015/#1016); re-enable on official v0.8.0 vectors")
```

**Suggested improvement:**
Gate the skip on a checked vector-version marker or revision. The test should run automatically when the vectors declare v0.8.0 or newer, and fail with a clear message if the version cannot be determined.

**Why:** This preserves the temporary OOM workaround without silently losing future end-to-end conformance coverage.

PR #1036, opened 2026-07-05
Feat/pvm-0.7.2-recompiler 

TwEricShen:
# feat(PVM): x86-64 recompiler (JIT) backend

Adds a second PVM execution backend — the **recompiler** — which compiles PVM basic
blocks to x86-64 native code and runs them directly (linux/amd64 + cgo, build-gated;
other platforms automatically fall back). The interpreter remains the ground truth;
the backend is selected via `JAM_PVM_BACKEND=recompiler` / `--pvm-backend recompiler`,
and the default is unchanged (interpreter).

Design details live in **`PVM/docs/`** (chapters 0–6, one per module). This
description is a map plus the verification results.

## Correctness verification

- Fuzz conformance — tiny dataset (0.7.2, 336 folders / 1179 JSON): **1179 PASSED / 0 FAILED** (including an eviction stress run with cache cap=1).
- Full dataset replay (fuzz-session capture, `.bin`): **all pass** (both interpreter and recompiler).
- PVMtrace per-instruction diff (interpreter vs recompiler): **800/800 identical** across all modes.
- `go test -race` (node fuzzy): 0 races.
- Unit tests: assembler encoding/label/fixup; compile+execute for all opcodes; signal handler (PROT_NONE / read-only / reserved zone / register preservation / coexistence with the Go runtime); sbrk; cache eviction (LRU / in-use protection).
- jam-testing **minifuzz** (forks, no_forks, fallback, safrole, storage, storage_light): **all pass under both backends**.
- jam-testing **picofuzz** (fallback, safrole, storage, storage_light): **all pass under both backends** (3 runs each; timing-only, responses not validated).

### picofuzz (tiny spec, median of 3).
Numbers are **per-block import latency** aggregated across all blocks in a suite (microseconds per block), not the suite's total wall-clock time. before = interpreter, after = recompiler:

| suite | mean before -> after | mean | p99 before -> after | p99 |
|---|---|---|---|---|
| fallback | 2554 -> 2572 us | +1% | 4352 -> 3893 us | -11% |
| safrole | 5971 -> 5979 us | +0% | 11433 -> 11943 us | +4% |
| storage | 6688 -> 3643 us | -46% | 21956 -> 13341 us | -39% |
| storage_light | 4164 -> 3406 us | -18% | 10961 -> 14123 us | +29% |

| suite | median before -> after | median | stdDev before -> after |
|---|---|---|---|
| fallback | 1878 -> 1901 us | +1% | 7550 -> 7200 us |
| safrole | 6878 -> 6875 us | -0% | 7885 -> 7861 us |
| storage | 4584 -> 3002 us | -35% | 8107 -> 7313 us |
| storage_light | 3481 -> 2837 us | -18% | 7362 -> 7403 us |

End-to-end whole-suite wall time (median of 3). Supporting signal only: most of it is fixed overhead (target container startup, per-block socket round-trips, debug logging), so it is noisier and dilutes the effect compared with the per-block numbers above:

| suite | wall time before -> after | change |
|---|---|---|
| fallback | 16.27 -> 16.28 s | +0% |
| safrole | 20.00 -> 19.94 s | -0% |
| storage | 20.75 -> 17.45 s | -16% |
| storage_light | 17.93 -> 17.24 s | -4% |

### test-timing-fuzz-trace
**Timing/bench**: per-ImportBlock STF timing in `internal/fuzz` —
  `make test-timing-fuzz-trace` (`scripts/run_fuzz_timing_test.sh`) ports
  `test-timing-jam-test-vectors-trace` to the fuzz target: `TIMING=1` +
  `PVM_BACKEND` server, `test_folder` replay, same STF summary on graceful
  shutdown. Plus `scripts/bench_fuzz_conformance.sh`, `scripts/run_pvmtrace_fuzz_capture.sh`
Per-run TOTAL (us/block): interpreter 28273 / 28522 / 28650, recompiler 22847 / 22921 / 23359.

| STF step | interpreter -> recompiler (avg us/block) | change |
|---|---|---|
| ValidateNonVRFHeader | 1561 -> 1507 us | -3% |
| UpdateBetaH | 2 -> 2 us | +5% |
| UpdateDisputes | 34 -> 32 us | -4% |
| UpdateSafrole | 16184 -> 16257 us | +0% |
| ValidateHeaderVrf | 1092 -> 1104 us | +1% |
| ValidateExtrinsic | 28 -> 28 us | +1% |
| UpdateAssurances | 287 -> 291 us | +2% |
| UpdateReports | 190 -> 186 us | -2% |
| **UpdateAccumulate** | **8934 -> 3452 us** | **-61%** |
| UpdateHistory | 17 -> 17 us | +1% |
| UpdatePreimages | 73 -> 80 us | +11% |
| UpdateAuthorizations | 3 -> 3 us | -0% |
| UpdateStatistics | 59 -> 64 us | +9% |
| **TOTAL** | **28482 -> 23042 us** | **-19%** |

## Datasets

All numbers in this PR reference one of two datasets:

| Name | Path | How it's run |
|------|------|--------------|
| **tiny** (conformance corpus) | `pkg/test_data/jam-conformance/fuzz-reports/0.7.2/traces/` — 336 folders / 1179 JSON | JSON fixtures, tiny chainspec: `make run-target` (server) + `go run ./cmd/fuzz/ test_folder <sock-path> <data-path>` |
| **full** (fuzz-session capture) | `session-26deee86e705508988baca0f952db5da/trace` | Binary fixtures, full chainspec: `JAM_FUZZ_SPEC=full go run ./cmd/fuzz/ test_folder --format=bin --skip 4 <sock-path> <data-path>` |

Flags for the full run: `JAM_FUZZ_SPEC=full` (the fuzzer needs the full chainspec
to decode `.bin` fixtures), `--format=bin` (fixture file format to read),
`--skip 4` (skip the first N blocks of the dataset, so `SetState` starts from a
later block).

### Execution core (`PVM/recompiler/`)
- **Sandbox**: a single mmap `[Control 4KB | Guest 4GB | Guard 4KB]` with R15
  addressing both ways (negative offsets = VM execution state, positive offsets =
  guest memory); all 13 PVM registers mapped to x86 registers with zero spills
  → `docs/0_Recompiler_Setup.md`
- **W^X dual-mapping code arena**: `memfd_create` + RW/RX views of the same
  physical pages — code is executable immediately after writing, zero `mprotect`
  anywhere → `docs/0`
- **Compiler**: `preDecodeBlocks` static decode → per-block lazy compilation →
  hand-written x86-64 assembler (REX/ModR/M/SIB encoding, integer label handles +
  fixup backpatching) → `docs/1_Recompiler_Workflow.md`, `docs/2_x86_Assembler.md`
- **Hardware memory protection**: SIGSEGV/SIGBUS/SIGFPE signal handler (C + TLS
  fault window) translates page faults into PVM `ExitPageFault`/`ExitPanic` —
  loads/stores need no software bounds checks → `docs/3_Signal_Handler.md`
- **Host calls**: `ecalli` → exit trampoline → Go omega dispatch → writeback;
  zero-copy guest memory (omega reads/writes the same mmap directly); sbrk is
  resolved at the machine layer → `docs/4_HostCall_Integration.md`
- **Native djump dispatch**: jump table + bitmask rodata + PC→native dispatch
  table; a hit never leaves native code → `docs/5_Djump_Dispatch.md`

### Performance (current state and methodology: `docs/TODO.md`, `docs/JIT_PROFILE.md`)
- **Native block chaining**: static exits (fallthrough / branch / jump) look up the
  dispatch table and `jmp` straight into the target block; misses self-heal after a
  single round-trip — Go↔native round-trips dropped **3.31M → 28.3k (117×)**, so
  loops converge to fully-native execution with essentially only the mandatory
  host-call exits remaining
- **Cross-invocation code cache**: compiled artifacts shared per CodeHash
  (single-flight), with LRU eviction (`JIT_CACHE_MAX_PROGRAMS`; refCount protects
  arenas that are still executing); the deblobbed `*Program` cache is shared by
  both backends
- **Codegen**: string labels → integer handles (compile time -47%); gas check fused
  from 4 instructions to 2 (`SUB`+`JS`, semantically equivalent per-instruction gas
  per GP 0.7.2, with the cold OOG path refunding the fused charge)
- Cumulative: total invoke time on the **tiny dataset** **~1897ms → ~1086ms**, and
  UpdateAccumulate is **1.57×** faster than the interpreter.
  On the **full dataset** the two backends are equally fast — PVM is <4% of STF there,
  and the system bottleneck is Safrole ring-VRF (not PVM; analysis in
  `docs/JIT_PROFILE_ANALYSIS.md`, fix listed in TODO)

### Observability and test tooling
- **PVMtrace** (`PVM/PVMtrace/`, `-tags trace`): per-instruction binary traces +
  first-divergence diff; `cmd/pvmtrace` CLI (`pvm-diff find-diff/show/detail`,
  deblob) → `docs/6_PVMtrace.md`, `docs/CONFORMANCE_TEST.md`
- **JIT_PROFILE** (env-gated, zero cost when off): per-phase timing +
  roundTrips/cache/host counters; `JIT_CPUPROFILE`/`JIT_PPROF_ADDR` pprof (the
  `phase=pvm` label isolates a PVM-only view); `JIT_PERFMAP` for per-block hotspots
  with Linux perf → `docs/JIT_PROFILE.md`
- **cmd/fuzz**: fixed the bug where the `--pvm-backend` default shadowed
  `JAM_PVM_BACKEND`; added `.bin` fixture support (`--format`/`--skip`)
- **Timing/bench**: per-ImportBlock STF timing in `internal/fuzz`
  (`make test-timing-fuzz-trace`), `scripts/bench_fuzz_conformance.sh`,
  `scripts/run_pvmtrace_fuzz_capture.sh`
- Docker: production and trace images + a recompiler unit-test environment
  (Makefile targets)

### Incidental fixes
- Interpreter invocation OOG check (`PVM/invocation.go`)
- PVM code restructure: the interpreter moved into `PVM/interpreter/`; both
  backends share decode, the `GuestMemory` interface, and the program cache

## Known limitations (`docs/TODO.md` §5)
- Page-crossing memory accesses have a theoretical semantic divergence from the
  interpreter in fault address / partial-write behavior (only when an access
  straddles a mapped/PROT_NONE boundary; conformance does not cover this detail)
- `jump_ind` + HALT `Counter == instr.PC` is aligned in code but still lacks a
  unit test

## Future work (see `docs/TODO.md`)
1. **JITContext pooling**: the per-invocation 4GB mmap/munmap (setup ~195ms) is now the largest remaining cost; move `SetFaultWindow` from per-block to per-invocation
2. Eviction follow-ups: proactive eviction, bytes-based cap, graceful arena-full
   handling
3. Block-based gas: `emitBlockGasCheck` is ready, waiting on GP 0.8.0 test vectors
4. (Non-PVM) ring-verifier cache: `GetVerifier` rebuilds account for ~45% of the
   full-dataset run; fixing it is a potential ~2× STF win

## Review guide
Suggested order: `docs/0` → `docs/1` (overall workflow) → `compiler.go` +
`emit_*.go` (translation rules) → `recompiler.go` / `execute.go` (execution loop)
→ `compiled_program.go` (cache lifecycle). To re-run conformance, see
`docs/CONFORMANCE_TEST.md`.


PVM/recompiler/host.go
```diff
@@ -0,0 +1,143 @@
...
+
+		input := PVM.OmegaInput{
+			Operation: PVM.OperationType(exitReason.GetHostCallID()),
+			VM:        &vm,
+			Addition:  h.Addition,
+			HostCalls: h.HostCalls,
+		}
+
+		omega := PVM.GetOmega(h.HostCalls, input.Operation)
+		PVMtrace.LogHostCallDispatchEnv(uint32(*h.Addition.ServiceID), PVM.HostCallName(int(input.Operation)), regsBuf, int64(*gasBuf))
```
yu2C: The recompiler host path unconditionally dereferences `h.Addition.ServiceID` when logging host-call dispatch, but `Psi_I` intentionally constructs `HostCallArgs` with `ServiceID = nil`. That means `is_authorized` can crash in the recompiler backend as soon as it reaches a host call, instead of returning a normal PVM panic / out-of-gas result.

Please guard this logging path for nil `ServiceID`, or only log the service identifier in invocation modes that guarantee it exists.

Makefile
```diff
@@ -133,31 +168,29 @@ release-target:
...
 	bash ./scripts/run_release.sh
 
-# --- Fuzz validation (spec: READMERef/VALIDATE_FUZZ.md) ---
-VALIDATE_FUZZ_SCRIPT := ./scripts/validate_fuzz.sh
-FUZZ_SMOKE_TRACE_DIR ?= 1766241814
-
-.PHONY: validate-fuzz validate-fuzz-ci validate-fuzz-vectors validate-fuzz-trace validate-fuzz-sock validate-fuzz-sock-smoke validate-fuzz-fuzzy validate-fuzz-jam-testing-local
-validate-fuzz:
-	$(VALIDATE_FUZZ_SCRIPT)
+# Build the recompiler unit-test container (linux/amd64)
```
yu2C, line 210: This change removes the existing `validate-fuzz*` Makefile entry points, but the repository still depends on them. `scripts/ci_fuzz_pre_push.sh` still calls `make validate-fuzz-ci`, and `READMERef/VALIDATE_FUZZ.md` still documents the same targets as the supported local validation interface.

Please restore the old `validate-fuzz*` targets as stable wrappers around `scripts/validate_fuzz.sh`, and keep the new timing / pvmtrace targets as additions rather than replacements.

docker/Dockerfile
```diff
@@ -1,56 +1,100 @@
...
-ARG GP_VERSION=
-ARG TARGET_VERSION=
-ARG GOOS=linux
-ARG GOARCH=amd64
+ARG GP_VERSION=latest
+ARG TARGET_VERSION=latest
 ARG OUTPUT=new-jamneration-target
+ARG BUILD_TAGS=
+
+RUN go mod tidy
```
yu2C: The new Dockerfile regresses Go dependency caching. We used to copy `go.mod` / `go.sum` and download modules before copying the full source tree, so source-only changes could reuse the module layer. The new order copies the entire repository first and only then runs `go mod tidy`, which invalidates the dependency step on every code change.

Suggested fix:
- copy `go.mod` / `go.sum` first
- run `go mod download`
- copy the full source tree afterwards
- avoid `go mod tidy` in the image build unless this PR intentionally changes the module graph

Makefile
```diff
@@ -91,15 +104,37 @@ fmt:
...
 	mkdir -p $(JAM_FUZZ_HOST_DIR)
-	JAM_FUZZ=1 JAM_FUZZ_SPEC=tiny JAM_FUZZ_DATA_PATH=$(JAM_FUZZ_HOST_DIR)/ JAM_FUZZ_SOCK_PATH=$(JAM_FUZZ_HOST_DIR)/fuzz.sock go run ./cmd/fuzz/
+	JAM_FUZZ=1 JAM_FUZZ_SPEC=tiny JAM_PVM_BACKEND=$(PVM_BACKEND) JAM_FUZZ_DATA_PATH=$(JAM_FUZZ_HOST_DIR)/ JAM_FUZZ_SOCK_PATH=$(JAM_FUZZ_HOST_DIR)/fuzz.sock go run ./cmd/fuzz/
 
 JAM_FUZZ_IMAGE ?= new-jamneration-target:latest
 # Matches CI release (linux/amd64). Required on arm64/aarch64 hosts (Apple Silicon, Linux ARM).
 DOCKER_PLATFORM ?= --platform linux/amd64
+JAM_FUZZ_TRACE_IMAGE ?= new-jamneration-target:trace
+
+.PHONY: fuzz-docker-build-trace
```
yu2C: `fuzz-docker-build-trace` does not currently build a trace-specific image consistently. It tags `$(JAM_FUZZ_IMAGE)` instead of `$(JAM_FUZZ_TRACE_IMAGE)`, and it does not pass `BUILD_TAGS=trace`.

Please make this target build and tag the trace image explicitly, and pass the trace build tag so the produced binary matches the target name.

Makefile
```diff
@@ -91,15 +104,37 @@ fmt:
 
 # Fuzz host dir (matches scripts/run_fuzz_target_docker.sh default bind-mount path on host).
 JAM_FUZZ_HOST_DIR ?= .jam_fuzz_docker_run
+PVM_BACKEND ?= recompiler
```
yu2C: `run-target` now defaults `PVM_BACKEND` to `recompiler`, but `cmd/fuzz` explicitly errors when the recompiler backend is not linked in the current build. That makes the default local workflow fail on non-`linux/amd64` environments instead of falling back to the interpreter.

Please keep the default portable, or only switch to `recompiler` when the build/runtime environment guarantees that backend is available.

YCC3741 (approved):

yu2C (approved):

